PR #18, opened 2022-01-03
Txm part 1

connorwstein:
- Backfilled some test coverage on the terra client
- Simplify using cosmos client for queries and sending txes as well as ws mgmt
- Added gas price querying from FCD node and the ability to specify a gas limit multiplier and fallback gas price in the spec
- Bug fix: ContractID needs to be ContractAddress in the wasm contract interactions
- Consistent receiver semantics on the terrra client (needs to use pointer semantics for the head tracker)
- Introduce a Start method for the client which boots the ws connection and starts the head tracker


pkg/terra/contract_config_tracker.go
```diff
@@ -31,43 +31,51 @@ func NewContractTracker(spec OCR2Spec, jobID string, client *Client, lggr Logger
...
-func (ct *Contract) Start() {
-	for {
-		select {
-		case data := <-ct.data:
-			// process received data
-			for _, event := range data.Events {
-				if event == "wasm-set_config" {
-					ct.newConfig <- struct{}{}
+func (ct *Contract) Start() error {
+	// TODO:; timeout
```
RyanRHall, line 48: this still needed?

pkg/terra/contract_config_tracker.go
```diff
@@ -31,43 +31,51 @@ func NewContractTracker(spec OCR2Spec, jobID string, client *Client, lggr Logger
...
 		}
-	}
+	}()
+	return nil
 }
 
 // Close stops the event channel listener loop and unsubscribes the job from the websocket
 func (ct *Contract) Close() error {
 	// stop listening loops
 	defer close(ct.stop)
```
RyanRHall, line 75: `Close()` returns before all listening loops are stopped. What do you think about adding a wait group so that the close is synchronous?

pkg/terra/contract_config_tracker.go
```diff
@@ -79,47 +87,39 @@ func (ct *Contract) Notify() <-chan struct{} {
...
 // LatestConfigDetails returns data by reading the contract state and is called when Notify is triggered or the config poll timer is triggered
 func (ct *Contract) LatestConfigDetails(ctx context.Context) (changedInBlock uint64, configDigest types.ConfigDigest, err error) {
-	// query state
-	queryParams := NewAbciQueryParams(ct.ContractID.String(), []byte(`"latest_config_details"`))
-
-	raw, err := ct.terra.Query(
-		ABCI,
-		[]interface{}{"custom/wasm/contractStore", queryParams},
-	)
+	queryParams := NewAbciQueryParams(ct.ContractAddress.String(), []byte(`"latest_config_details"`))
```
RyanRHall, line 90: :raised_hands:

RyanRHall (changes requested):

pkg/terra/contract_config_tracker.go
connorwstein: yeah, there's a couple places where we need timeouts. I'll just add them now

pkg/terra/contract_config_tracker.go
```diff
@@ -31,43 +31,51 @@ func NewContractTracker(spec OCR2Spec, jobID string, client *Client, lggr Logger
...
-		select {
-		case data := <-ct.data:
-			// process received data
-			for _, event := range data.Events {
-				if event == "wasm-set_config" {
-					ct.newConfig <- struct{}{}
+func (ct *Contract) Start() error {
+	// TODO:; timeout
+	txes, err := ct.terra.clientCtx.Client.Subscribe(context.TODO(), ct.JobID,
+		fmt.Sprintf("tm.event='Tx' AND execute_contract.contract_address='%s'", ct.ContractAddress))
```
aalu1418, line 50: So it seems like the WS is used for the head tracker and catching new set config events...we could reduce the number of connections by removing it.
* libocr will poll for the configs anyway (so the WS isn't really necessary)
* we could also replace the head tracker with just reading from rpc or another endpoint

it would reduce the complexity of having to subscribe and unsubscribe

pkg/terra/relay.go
```diff
@@ -32,9 +34,14 @@ type OCR2Spec struct {
 	IsBootstrap bool
 
 	// network data
-	NodeEndpointHTTP string
-	NodeEndpointWS   string
-	ChainID          string
+	TendermintURL      string // URL exposing tendermint RPC (default port is 26657)
+	CosmosURL          string // URL exposing cosmos endpoints (port is 1317, needs to be enabled in terra node config)
```
aalu1418, line 38: are these WS urls or just http/rpc urls?

aalu1418 (commented):
🎉 amazing on the simplification of the terra blockchain connection code

pkg/terra/relay.go
connorwstein: just http/rpc

pkg/terra/contract_config_tracker.go
connorwstein: ah yes that did cross my mind, but the ws support basically just came for free with the cosmos client so it was quicker to actually drop it in as is. Will take a look to see if I can swap it out now

pkg/terra/relay.go
aalu1418: oh interesting! they don't require an actual WS connection to "subscribe"

pkg/terra/contract_config_tracker.go
aalu1418: up to you! i thought it originally would be an additional endpoint, but it seems like we get subscriptions for free. i'm curious how they do it - polling? or maybe something else

...but i guess removing the subscribing parts would reduce the number of goroutines + sub/unsub pieces

pkg/terra/relay.go
aalu1418: did a little digging and I think it does use the WS connection if we use subscribe...it might be appending `/websocket` to the url so then it can access it for subscriptions

https://github.com/cosmos/cosmos-sdk/blob/33dbf6a795e68c744a2c312b76d41614b3563c01/client/utils.go#L79
https://pkg.go.dev/github.com/tendermint/tendermint@v0.34.15/rpc/client/http#New

pkg/terra/contract_config_tracker.go
connorwstein: on second thought was able to just remove ws all together

pkg/terra/contract_config_tracker.go
connorwstein: ah yep easy, done

pkg/terra/client.go
```diff
@@ -56,66 +54,29 @@ func NewClient(spec OCR2Spec, lggr Logger) (*Client, error) {
 		WithTxConfig(app.MakeEncodingConfig().TxConfig)
 
 	return &Client{
-		close:              make(chan struct{}),
 		codec:              codec.NewLegacyAmino(),
 		chainID:            spec.ChainID,
 		clientCtx:          clientCtx,
 		cosmosRPC:          spec.CosmosURL,
-		httpClient:         &http.Client{Timeout: spec.HTTPTimeout},
```
aalu1418: nit:
why get rid of the client? it seems cleaner than recreating the contexts each time?

aalu1418 (approved):
LGTM! let's give it a go in local testing 🚀

pkg/terra/client.go
connorwstein: I'm thinking there might be other non-http.Client places we want to use that timeout and pulling it from an already constructed http.Client feels weird

aalu1418 (approved):
🚀

RyanRHall (approved):

PR #22, opened 2022-01-10
go test -race in CI

topliceanu:


.github/workflows/relay.yml
```diff
@@ -46,6 +46,9 @@ jobs:
       - name: Run unit tests
         run: nix develop -c go test -v ./pkg/terra/...
 
+      - name: Run tests with the race detector enabled
+        run: nix develop -c go test -v ./pkg/terra/... -race -count=10
```
jmank88, line 50: Could consider using the `GORACE` env var to isolate the failures from the verbose test output:
```suggestion
        run: GORACE="log_path=$PWD/race" nix develop -c go test -v ./pkg/terra/... -race -count=10
```

PR #23, opened 2022-01-11
Fix schema generation CI

RodrigoAD:


.github/workflows/contracts-release.yml
```diff
@@ -16,8 +16,8 @@ jobs:
           --mount type=volume,source="$(basename "$(pwd)")_cache",target=/code/target \
           --mount type=volume,source=registry_cache,target=/usr/local/cargo/registry \
           cosmwasm/workspace-optimizer:0.12.3
-      - name: Build Schemas
-        run: for d in contracts/*/; do     cd "$d"; cargo schema; cd ../..; done
+      # - name: Build Schemas
```
krebernisak, line 19: We'll merge this PR for now to unblock progress, but extracting this as a GH Issue so we can revisit it soon.

Also, note that any releases we do in between will need to be redone (missing schema) after the fix is out.

krebernisak (approved):

.github/workflows/contracts-release.yml
krebernisak: GH Issue:

- https://github.com/smartcontractkit/chainlink-terra/issues/30

PR #26, opened 2022-01-13
Txm part 3

connorwstein:
- Expose gas params in the writer api 
- Nil pointer fix in contract store
- Use logs for parsing (it appears not all nodes support events)
- Include batch simulation

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,7 +8,7 @@ require (
 	github.com/pelletier/go-toml v1.9.4
 	github.com/pkg/errors v0.9.1
 	github.com/satori/go.uuid v1.2.0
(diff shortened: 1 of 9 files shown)


go.mod
```diff
@@ -4,10 +4,11 @@ go 1.17
 
 require (
 	github.com/cosmos/cosmos-sdk v0.44.5
+	github.com/cosmos/cosmos-sdk/errors v1.0.0-beta.2
```
jmank88, line 7: Isn't nesting `go.mod`s in a repo (other than `v*/` dirs) an anti-pattern? We're stuck though?

connorwstein: fixed with go mod tidy

jmank88: bonus

jmank88 (approved):

PR #29, opened 2022-01-13
support multiple chains via ChainSet

jmank88:


pkg/terra/relay.go
```diff
@@ -34,9 +32,12 @@ type MsgEnqueuer interface {
 
 // CL Core OCR2 job spec RelayConfig member for Terra
 type RelayConfig struct {
+	TerraChainID string
```
connorwstein, line 35: nit: do we need the terra prefix since this is in the terra package?

connorwstein (dismissed):

pkg/terra/relay.go
jmank88: Not necessary, but could improve readability depending on context and avoids confusion about the data type :imp: :avocado: :shrug: 
We do this with `EVMChainID`/`evm_chain_id` in other places where it's not necessary, but actually _not_ on the evm `RelayConfig`, which this should probably be aligned with.

jmank88: I'm thinking remove it...

RyanRHall (dismissed):

connorwstein (approved):

PR #33, opened 2022-01-18
fix: reorder transmitter not needed for bootstrap

aalu1418:
Moved transmitter parsing to allow bootstrap to be created without transmitter key


krebernisak (approved):
🚀

krebernisak (approved):

PR #34, opened 2022-01-18
convert chain config to interface

jmank88:
- Convert to an interface for live reconfiguration, and add four more fields.
- Move some types from core

### DIFF
--- a/pkg/terra/config.go
+++ b/pkg/terra/config.go
@@ -13,9 +13,8 @@ var DefaultConfigSet = ConfigSet{
 	// ~8s per block, so ~80s until we give up on the tx getting confirmed
 	// Anecdotally it appears anything more than 4 blocks would be an extremely long wait.
 	BlocksUntilTxTimeout:  10,
-	ConfirmMaxPolls:       100,
 	ConfirmPollPeriod:     time.Second,
-	FallbackGasPriceULuna: sdk.MustNewDecFromStr("0.01"),
+	FallbackGasPriceULuna: sdk.MustNewDecFromStr("0.015"),
 	GasLimitMultiplier:    1.5,
 	// The max gas limit per block is 1_000_000_000
 	// https://github.com/terra-money/core/blob/d6037b9a12c8bf6b09fe861c8ad93456aac5eebb/app/legacy/migrate.go#L69.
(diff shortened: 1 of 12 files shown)


pkg/terra/db/db.go
```diff
@@ -0,0 +1,84 @@
...
+	// Valid next states, none, terminal state
+	Errored State = "errored"
+)
+
+type Msg struct {
+	ID         int64
+	ChainID    string `db:"terra_chain_id"`
+	ContractID string
+	State      State
+	Raw        []byte
```
jmank88, line 80: Renamed from `Msg` to avoid nesting, does `Raw` make sense?

pkg/terra/config.go
```diff
@@ -0,0 +1,143 @@
...
+	"github.com/smartcontractkit/chainlink-terra/pkg/terra/db"
+)
+
+var DefaultConfigSet = ConfigSet{
+	// ~8s per block, so ~80s until we give up on the tx getting confirmed
+	// Anecdotally it appears anything more than 4 blocks would be an extremely long wait.
+	BlocksUntilTxTimeout:  10,
+	ConfirmMaxPolls:       100,
+	ConfirmPollPeriod:     time.Second,
+	FallbackGasPriceULuna: sdk.MustNewDecFromStr("0.01"),
```
connorwstein, line 18: See comment here https://github.com/smartcontractkit/chainlink-terra/pull/32/files#r787938891 on whether we want to keep this. Also we should increase it to 0.015

pkg/terra/config.go
```diff
@@ -0,0 +1,143 @@
...
+	// we could include 1000 msgs which use up to 1M gas.
+	// To be conservative and since the number of messages we'd
+	// have in a batch on average roughly correponds to the number of terra ocr jobs we're running (do not expect more than 100),
+	// we can set a max msgs per batch of 100.
+	MaxMsgsPerBatch: 100,
+}
+
+type Config interface {
+	BlocksUntilTxTimeout() int64
+	ConfirmMaxPolls() int64
```
connorwstein, line 34: actually I think we can simplify here since max polls can be derived from blocks until tx timeout and poll period. We can just specify those two then max polls = timeout blocks * time per block (upper bound) / poll period

pkg/terra/db/db.go
connorwstein: yep, maybe just comment that its a serialized msg

connorwstein (approved):

pkg/terra/median_contract.go
```diff
@@ -18,9 +18,7 @@ import (
 	"github.com/smartcontractkit/libocr/offchainreporting2/types"
 )
 
-const (
-	BlockRate = 5 // 1 block/5 seconds
-)
+const BlockRateSeconds = 8 // 1 block/8 seconds
```
krebernisak, line 21: Why don't we extract this to the config, so it's a little less static than this?

Bombay-12 testnet looks to produce blocks every ~6s by my investigation.

krebernisak (dismissed):
Looks great!

connorwstein (approved):

krebernisak (approved):

PR #36, opened 2022-01-20
SetConfig optimization + billing config improvements

archseer:
Need to fit each call into 4kb

[please don't squash]


contracts/ocr2/src/contract.rs
```diff
@@ -73,6 +73,9 @@ pub fn instantiate(
         billing: Billing {
             recommended_gas_price: 0,
             observation_payment: 0,
+            base_gas: None,
+            gas_per_signature: None,
+            gas_adjustment: None,
```
archseer, line 78: These are now settable via `setBilling` but can also be `nil` to use the built-in default

contracts/ocr2/src/contract.rs
```diff
@@ -1196,18 +1196,21 @@ pub fn query_oracle_observation_count(deps: Deps, transmitter: String) -> StdRes
...
+    config: &Billing,
     juels_per_luna: u128,
     signature_count: usize,
 ) -> Uint128 {
-    const BASE_GAS: Decimal = decimal(338_000);
-    const GAS_PER_SIGNATURE: Decimal = decimal(67_000);
+    let signature_count = decimal(signature_count as u64);
+    let gas_per_signature = decimal(config.gas_per_signature.unwrap_or(17_000));
+    let base_gas = decimal(config.base_gas.unwrap_or(84_000));
+    let gas_adjustment = Decimal::percent(u64::from(config.gas_adjustment.unwrap_or(140)));
```
archseer, line 1206: Adjusted based on devnet values. We seemed to over-estimate by almost exactly 4x. The amount charged is also the "gas allocated", not "gas used" and the difference seems to be about 1.4x

contracts/ocr2/src/msg.rs
```diff
@@ -29,25 +29,25 @@ pub struct InstantiateMsg {
...
-        signers: Vec<Vec<u8>>,
+        signers: Vec<Binary>,
         transmitters: Vec<String>,
         f: u8,
-        onchain_config: Vec<u8>,
+        onchain_config: Binary,
         offchain_config_version: u64,
-        offchain_config: Vec<u8>,
+        offchain_config: Binary,
     },
```
archseer, line 38: @RodrigoAD all of these are now base64 encoded to save space. Depending on the library you're using it might need some changes

contracts/ocr2/src/contract.rs
```diff
@@ -73,6 +73,9 @@ pub fn instantiate(
         billing: Billing {
             recommended_gas_price: 0,
             observation_payment: 0,
+            base_gas: None,
```
RodrigoAD, line 76: Do we need `transmission_payment`?

archseer: Good point, looks like that is missing altogether. I'll add it in 👍🏻

archseer: Added

pkg/terra/contract_transmitter.go
```diff
@@ -55,7 +56,7 @@ func (ct *ContractTransmitter) Transmit(
 	}
 	msgStruct.Transmit.Report = []byte(report)
 	for _, sig := range sigs {
-		msgStruct.Transmit.Signatures = append(msgStruct.Transmit.Signatures, sig.Signature)
+		msgStruct.Transmit.Signatures = append(msgStruct.Transmit.Signatures, []byte(base64.StdEncoding.EncodeToString(sig.Signature)))
```
krebernisak, line 59: Another line here would help readability:

```suggestion
                sigBase64 := []byte(base64.StdEncoding.EncodeToString(sig.Signature))
		msgStruct.Transmit.Signatures = append(msgStruct.Transmit.Signatures, sigBase64)
```

contracts/ocr2/src/msg.rs
krebernisak: Let's reuse the same PR to add any Gauntlet changes.

contracts/ocr2/src/integration_tests.rs
```diff
@@ -439,23 +441,27 @@ fn transmit_happy_path() {
...
-        f: 1,
-        onchain_config: vec![],
+        f: 6,
+        onchain_config: Binary(vec![]),
         offchain_config_version: 2,
-        offchain_config: vec![5, 5, 6],
+        offchain_config: Binary(vec![1; 2165]),
     };
+
+    const MAX_MSG_SIZE: usize = 4 * 1024; // 4kb
```
krebernisak, line 462: We should find out how much overhead does Serum mutlisig exactly adds, and consider it here while testing the tx size limit with the max oracles number. 

@ebarakos can you please research this and add it in?

contracts/ocr2/src/state.rs
```diff
@@ -35,12 +35,17 @@ pub struct Validator {
     pub gas_limit: u64,
 }
 
-#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
+#[derive(Serialize, Deserialize, Clone, Default, Debug, PartialEq, JsonSchema)]
 pub struct Billing {
     /// Should match https://fcd.terra.dev/v1/txs/gas_prices
     /// NOTE: needs to be scaled to the same amount of decimals places as LINK token
```
krebernisak, line 41: Which configs here need to be scaled, please be a little more specific here.

Can we do a quick and safe sanity check on these configs on the input so we revert if:

- values are really low (not scaled properly)?
- values are under contract defaults, does it make sense to go under defaults?

contracts/ocr2/src/integration_tests.rs
krebernisak: If we go with the "approvable proposal by digest" multisig proposal, then this suggestion won't be relevant anymore.

contracts/ocr2/src/state.rs
```diff
@@ -35,12 +35,17 @@ pub struct Validator {
...
     pub recommended_gas_price: u64,
     pub observation_payment: u64,
+    pub transmission_payment: u64,
+    pub base_gas: Option<u64>,
```
krebernisak, line 45: I'd suggest we rename this as `gas_base` to be consistent with other args.

contracts/ocr2/src/state.rs
krebernisak: Extracted as GH Issue:

- https://github.com/smartcontractkit/chainlink-terra/issues/58

contracts/ocr2/src/state.rs
```diff
@@ -35,12 +35,17 @@ pub struct Validator {
     pub gas_limit: u64,
 }
 
-#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
+#[derive(Serialize, Deserialize, Clone, Default, Debug, PartialEq, JsonSchema)]
 pub struct Billing {
     /// Should match https://fcd.terra.dev/v1/txs/gas_prices
     /// NOTE: needs to be scaled to the same amount of decimals places as LINK token
     pub recommended_gas_price: u64,
```
krebernisak, line 42: Suggestion by @archseer:

> for recommended_gas_price: the API returns uLUNA but we want to turn that into a u128 with 18 decimal places
> (uluna = micro luna = luna * 10⁻⁶)
>
> there wasn't an easy way to construct a decimal and say "this only has 6 decimal places" because it hardcodes to 18 decimals https://github.com/smartcontractkit/chainlink-terra/blob/9c529e3c016e23234cdbde8b4d9c7952ebe2912f/contracts/ocr2/src/decimal.rs#L30
>
> well there's https://github.com/smartcontractkit/chainlink-terra/blob/9c529e3c016e23234cdbde8b4d9c7952ebe2912f/contracts/ocr2/src/decimal.rs#L76-L79 so we could technically support uluna

krebernisak: GH Issue:

- https://github.com/smartcontractkit/chainlink-terra/issues/59

contracts/ocr2/schema/execute_msg.json
```diff
@@ -364,9 +334,35 @@
       "type": "object",
       "required": [
         "observation_payment",
-        "recommended_gas_price"
+        "recommended_gas_price",
```
krebernisak, line 337: These are stale and should be regenerated.

contracts/ocr2/src/contract.rs
```diff
@@ -929,11 +931,15 @@ pub fn execute_set_billing(
         Event::new("set_billing")
             .add_attribute(
                 "recommended_gas_price",
```
krebernisak, line 933: Should we rename this attribute as `recommended_gas_price_uluna` to keep it consistent?

contracts/ocr2/src/contract.rs
```diff
@@ -969,10 +975,12 @@ pub fn execute_withdraw_payment(
 fn owed_payment(config: &Config, transmitter: &Transmitter) -> StdResult<Uint128> {
     let rounds = config.latest_aggregator_round_id - transmitter.from_round_id;
 
-    Ok(Uint128::from(config.billing.observation_payment)
-        .checked_mul(rounds.into())?
-        // + transmitter gas reimbursement
-        .checked_add(transmitter.payment)?)
+    Ok(
+        Uint128::new(u128::from(config.billing.observation_payment_gjuels) * 10u128.pow(9)) // scale to juels
```
krebernisak, line 979: Consider extracting this scale factor `10u128.pow(9)` as a constant.

jmank88:
@archseer There are two more places signers are handled as <Vec<Vec<u8>> still:
https://github.com/smartcontractkit/chainlink-terra/blob/c5e735e9fae73580d330bfde74db64c11bcf9a39/contracts/ocr2/src/contract.rs#L109
https://github.com/smartcontractkit/chainlink-terra/blob/c5e735e9fae73580d330bfde74db64c11bcf9a39/contracts/ocr2/src/contract.rs#L235

archseer:
That's converting/unwrapping from the `Binary` type, internally we just use byte vecs https://github.com/smartcontractkit/chainlink-terra/blob/c5e735e9fae73580d330bfde74db64c11bcf9a39/contracts/ocr2/src/msg.rs#L30-L32

packages-ts/gauntlet-terra-contracts/src/commands/contracts/ocr2/setBilling.ts
```diff
@@ -23,15 +24,16 @@ const makeCommandInput = async (flags: any, args: string[]): Promise<CommandInpu
...
 }
 
 const makeContractInput = async (input: CommandInput): Promise<ContractInput> => {
   return {
     config: {
-      recommended_gas_price: new BN(input.recommendedGasPrice).toNumber(),
-      observation_payment: new BN(input.observationPaymentGjuels).toNumber(),
+      observation_payment_gjuels: new BN(input.observationPaymentGjuels).toNumber(),
+      transmission_payment_gjuels: new BN(input.transmissionPaymentGjuels).toNumber(),
+      recommended_gas_price_uluna: input.recommendedGasPriceUluna
```
RodrigoAD, line 36: This should be of Decimal type (https://github.com/smartcontractkit/chainlink-terra/pull/36/files#diff-7593d1da79e988b58adce267ba96df185c0abebbfc17323f0de5e749e6419134R56)

@archseer Which format should we be applying here?

archseer: It's a string, you can pass through the value from https://fcd.terra.dev/v1/txs/gas_prices without modifying it

RodrigoAD: This info comes from the RDD, so we can expect it to come in the right format

jmank88 (approved):

packages-ts/gauntlet-terra-contracts/src/commands/contracts/ocr2/setBilling.ts
```diff
@@ -33,7 +33,7 @@ const makeContractInput = async (input: CommandInput): Promise<ContractInput> =>
     config: {
       observation_payment_gjuels: new BN(input.observationPaymentGjuels).toNumber(),
       transmission_payment_gjuels: new BN(input.transmissionPaymentGjuels).toNumber(),
-      recommended_gas_price_uluna: input.recommendedGasPriceUluna
+      recommended_gas_price_uluna: new BN(input.recommendedGasPriceUluna).toString(),
```
reductionista, line 36: If this works, I think it will have the effect of converting the decimal an integer, either by truncating or rounding a suppose.  Is that really what we want to do?

From:  https://github.com/indutny/bn.js/
"Note: decimals are not supported in this library."

RodrigoAD: You're completely right. I'll open a fix

archseer: Looks like it's handled in https://github.com/smartcontractkit/chainlink-terra/pull/76

PR #42, opened 2022-01-21
poll for and cache contract state

jmank88:


pkg/terra/median_contract.go
```diff
@@ -50,6 +152,29 @@ func (ct *MedianContract) LatestTransmissionDetails(
...
 	latestTimestamp time.Time,
 	err error,
+) {
+	ct.trnMu.Lock()
+	configDigest = ct.trn.configDigest
+	epoch = ct.trn.epoch
+	round = ct.trn.round
+	latestAnswer = ct.trn.latestAnswer
+	latestTimestamp = ct.trn.latestTimestamp
+	ct.trnMu.Unlock()
```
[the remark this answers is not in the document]
aalu1418: A follow up to this...if they are supposed to have `R` what's the difference in use between `LatestTransmissionDetails` and `latestTransmissionDetails`?

aalu1418 (commented):
some preliminary thoughts 😄

pkg/terra/contract_transmitter.go
```diff
@@ -37,6 +53,63 @@ func NewContractTransmitter(jobID string,
...
+		ct.lggr.Debugf("Stopping")
+		close(ct.stop)
+		<-ct.done
+		return nil
+	})
+}
+
+func (ct *ContractTransmitter) pollState() {
+	defer close(ct.done)
+	tick := time.After(utils.WithJitter(ct.cfg.OCRCachePollPeriod()))
```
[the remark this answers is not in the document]
jmank88: Or syncing up with the block period for that matter.

pkg/terra/median_contract.go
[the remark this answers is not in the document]
jmank88: Whoops. Bad copy/paste work.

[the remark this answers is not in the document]
jmank88: Do you mean latest transmission vs. round? I think there's probably an optimization there, yeah.

pkg/terra/contract_config_tracker.go
```diff
@@ -3,41 +3,140 @@ package terra
...
+	defer close(ct.done)
+	tick := time.After(0)
+	for {
+		select {
+		case <-ct.stop:
+			return
+		case <-tick:
+			ctx, cancel := utils.ContextFromChan(ct.stop)
+			done := make(chan struct{})
+			go func() {
```
connorwstein, line 79: this goroutine is just to work around the fact that the ctx cancellation isn't respected right? I think we could still just do it in line since the latestConfigDetails call is still bounded by the http timeout. It would mean that shutting down could potentially take the length of the http timeout in the worse case, but that doesn't seem so bad actually and would ensure we never leak any goroutines and keep things simpler

pkg/terra/median_contract.go
```diff
@@ -29,18 +31,92 @@ type LatestConfigReader interface {
...
+func (ct *MedianContract) Close() error {
+	return ct.StopOnce("TerraMedianContract", func() error {
+		ct.lggr.Debugf("Stopping")
+		close(ct.stop)
+		<-ct.done
+		return nil
+	})
+}
+
+func (ct *MedianContract) pollState() {
```
connorwstein, line 81: yeah I think it would be cleaner to combine these pollers into a single state cache object like you mentioned. It would get injected into MedianContract/ContractTracker/ContractTransmitter etc. who accept an interface describing whatever they need from the cache. Just a single goroutine + for/select loop would be make it easier to reason about the rpc call frequency. Wdyt?

pkg/terra/contract_config_tracker.go
jmank88: Yeah I don't like leaking goroutines, but I think we're stuck with that as the less bad choice, since it's critical to yield for shutdown to proceed properly. I do think it can be done more simply without this extra level though :+1: . Fortunately there aren't any external side effects, since they just update internal state.

pkg/terra/median_contract.go
krebernisak: Second this!

pkg/terra/contract_cache.go
```diff
@@ -0,0 +1,426 @@
...
+	configDigest types.ConfigDigest,
+	epoch uint32,
+	round uint8,
+	err error,
+) {
+	cc.transMu.RLock()
+	ts := cc.transTS
+	configDigest = cc.digest
+	epoch = cc.epoch
+	round = cc.round
```
jmank88, line 401: The Solana version returns `0`s here for epoch and round. Should we do the same?

jmank88:
Update with `ContractCache` to do the heavy lifting, and manage a single polling loop with three serial queries for config (x2) and transmission.

pkg/terra/contract_cache.go
connorwstein: no, that actually seems like a bug in the solana version or possibly they are intentionally not supporting the explicit round requesting for some reason

jmank88: @aalu1418  https://github.com/smartcontractkit/chainlink-solana/blob/develop/pkg/solana/median_contract.go#L59

connorwstein: yeah turns out for solana the rounds are fast enough that they reasoned to just disable this feature essentially

pkg/terra/contract_cache.go
```diff
@@ -0,0 +1,428 @@
...
+
+	cosmosSDK "github.com/cosmos/cosmos-sdk/types"
+
+	"github.com/smartcontractkit/chainlink/core/utils"
+	"github.com/smartcontractkit/libocr/offchainreporting2/types"
+
+	"github.com/smartcontractkit/chainlink-terra/pkg/terra/client"
+)
+
+type ContractCache struct {
```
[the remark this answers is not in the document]
jmank88: Much cleaner!

pkg/terra/contract_reader.go
```diff
@@ -0,0 +1,261 @@
...
+
+func NewOCR2Reader(addess cosmosSDK.AccAddress, chainReader client.Reader, lggr Logger) *OCR2Reader {
+	return &OCR2Reader{
+		address:     addess,
+		chainReader: chainReader,
+		lggr:        lggr,
+	}
+}
+
+func (or *OCR2Reader) fetchLatestConfigDetails(ctx context.Context) (changedInBlock uint64, configDigest types.ConfigDigest, err error) {
```
krebernisak, line 33: I'd suggest we make these functions public and drop the `fetch*` prefix. The reader could be useful for other components of the system (monitoring, test assertions, etc.).

```suggestion
func (r *OCR2Reader) LatestConfigDetails(ctx context.Context) (changedInBlock uint64, configDigest types.ConfigDigest, err error) {
```

jmank88: I did the same with `ContractCache`, and embedded them as well to eliminate the boilerplate method forwarding (and the whole of `MedianContract` - the cache implements the interface directly now). Much simpler overall, but happy to switch back to explicit forwarding if embedding is confusing or error prone.

connorwstein (approved):

PR #48, opened 2022-01-25
add optional RelayConfig.NodeName; move fcd url to chain config

jmank88:


pkg/terra/relay.go
```diff
@@ -32,11 +32,13 @@ type MsgEnqueuer interface {
...
 // CL Core OCR2 job spec RelayConfig member for Terra
 type RelayConfig struct {
-	ChainID string `json:"chainID"`
+	ChainID  string `json:"chainID"`  // required
+	NodeName string `json:"nodeName"` // optional
 }
 
 type OCR2Spec struct {
-	ChainID string
+	ChainID  string
```
krebernisak, line 40: We should either embed the `RelayConfig` or add it as this is how the spec is actually modeled.

pkg/terra/relay.go
```diff
@@ -32,11 +32,13 @@ type MsgEnqueuer interface {
 
 // CL Core OCR2 job spec RelayConfig member for Terra
 type RelayConfig struct {
-	ChainID string `json:"chainID"`
+	ChainID  string `json:"chainID"`  // required
+	NodeName string `json:"nodeName"` // optional
```
krebernisak, line 36: Why is `NodeName` still optional here? b/c we'll still use the first node as the one if not specified, right?

Hmm, maybe we should also add the `chain-id` to the `node` data structure so we have that link and it can be verified. If the node is not explicitly specified we use a random (or first) one compatible with the specified chain.

If we don't know the `chain-id` for the `node`, and we leave the `spec.relay.nodeName` optional there is room for errors by misconfiguration.

pkg/terra/client/gas_price_estimator.go
```diff
@@ -50,20 +50,17 @@ func (gpe *ClosureGasPriceEstimator) GasPrices() (map[string]sdk.DecCoin, error)
...
 
-func NewFCDGasPriceEstimator(fcdURLRaw string, requestTimeout time.Duration, lggr Logger) (*FCDGasPriceEstimator, error) {
-	// Sanity check the URL works and populate the cached value
-	fcdURL, err := url.Parse(fcdURLRaw)
-	if err != nil {
-		return nil, err
-	}
+type Config interface{ FCDURL() url.URL }
+
+func NewFCDGasPriceEstimator(cfg Config, requestTimeout time.Duration, lggr Logger) *FCDGasPriceEstimator {
```
krebernisak, line 60: Why send full config in when we just use the URL? Makes users go through additional hoops to prepare the struct, which seem unnecessary.

pkg/terra/config.go
```diff
@@ -60,16 +96,16 @@ type ConfigSet struct {
...
+	defaults configSet
 	chain    db.ChainCfg
 	chainMu  sync.RWMutex
 	lggr     Logger
 }
 
-// NewConfig returns a Config with defaults overridden by dbcfg.
-func NewConfig(dbcfg db.ChainCfg, defaults ConfigSet, lggr Logger) *config {
+// NewConfig returns a Config with defaults for id overridden by dbcfg.
+func NewConfig(id string, dbcfg db.ChainCfg, lggr Logger) *config {
```
krebernisak, line 106:
```suggestion
func NewConfig(chainID string, dbcfg db.ChainCfg, lggr Logger) *config {
```

krebernisak: Or perhaps even type it explicitly as `id ChainID`, then move the `defaultsForID` method to the new `ChainID` type as `defaults`.

pkg/terra/relay.go
jmank88: `chain-id` is already wired through, and the core PR updates the name-less case from _first_ to _random_ (still limited by chain id) to try to distribute any load by default: https://github.com/smartcontractkit/chainlink/pull/5853/files#diff-10e1587402f9125b6af71481bf4247044f5b411800c7367f091b7e4f91a64ad3R107

pkg/terra/client/gas_price_estimator.go
jmank88: The config is dynamic, so this will now pick up runtime reconfiguration.

pkg/terra/relay.go
krebernisak: Aha yeah, `chain-id` is already wired. Oh, this is nice then :) 

I'll document it in the job-spec example I'm updating...

pkg/terra/client/gas_price_estimator.go
krebernisak: Aha, the interface serves as a pointer to the latest config. Got it!

krebernisak: Maybe additionally consider adding a relevant comment somewhere, so it's clear why we pass in config interfaces like this.

pkg/terra/relay.go
krebernisak: Please just extend the `// optional` comment with the same info here.

jmank88: Please do try to break this in testing though :wink: 
All of the Terra parts in core should be cleanly separated by chain id like with EVM: chainSet, nodes, txm, etc.

krebernisak (dismissed):

pkg/terra/config.go
jmank88: I like the clarity of the helper, but using it for the parameter type is a bit of a burden for the callers since conversion has to be explicit.

krebernisak: Yup, crossed my mind. I'm good either way for now. Consider doing what feels better to you.

pkg/terra/config.go
```diff
@@ -1,15 +1,52 @@
...
+
+// defaults returns Chain specific defaults.
+func (id ChainID) defaults() configSet {
+	c := defaultConfigSet
+	lower := strings.ToLower(string(id))
+	switch {
+	case strings.HasPrefix(lower, "columbus-"):
+		c.FCDURL = *mainnetFCDURL
+	case strings.HasPrefix(lower, "bombay-"):
+		c.FCDURL = *bombayFCDURL
```
jmank88, line 43: Perhaps this could default to mainnet for convenience? Could also be risky though, if it surprises someone.

pkg/terra/client/gas_price_estimator.go
```diff
@@ -91,20 +90,24 @@ type prices struct {
 }
 
 func (gpe *FCDGasPriceEstimator) request() (map[string]sdk.DecCoin, error) {
-	req, _ := http.NewRequest(http.MethodGet, gpe.fcdURL.String(), nil)
+	fcdURL := gpe.cfg.FCDURL()
```
jmank88, line 93: If we don't have a global default fallback for this, then this should probably sanity check if it's empty and return a clear misconfiguration error.

connorwstein: can we do the sanity check at chain add time? it's be cleaner if the config interface just always returns non-empty values

pkg/terra/config.go
connorwstein: hm we already have a fallback gas price, so I'm not sure we also need fallback fcd urls. Could we just check that the URL they provide is a real url at chain add time? Then if its providing invalid data we fallback to our fallback gas price

jmank88: Yeah that makes sense :+1:

connorwstein (approved):

PR #53, opened 2022-01-26
add version input and fetch contract bytecode from Github release

akhilchainani:


packages-ts/gauntlet-terra-contracts/src/commands/tooling/upload.ts
```diff
@@ -44,7 +45,8 @@ export default class UploadContractCode extends TerraCommand {
     const contractReceipts = {}
     const responses: any[] = []
     for (const contractName of askedContracts) {
-      const contract = getContract(contractName)
+      const version = this.flags.version ? this.flags.version : DEFAULT_RELEASE_VERSION
+      const contract = getContract(contractName, this.flags.version)
```
RodrigoAD, line 49: `const contract = getContract(contractName, version)`

packages-ts/gauntlet-terra-contracts/src/commands/abstract/index.ts
```diff
@@ -20,12 +21,12 @@ export const makeAbstractCommand = async (
   args: string[],
   input?: any,
 ): Promise<TerraCommand> => {
-  const commandOpts = parseInstruction(instruction)
+  const commandOpts = parseInstruction(instruction, flags)
```
RodrigoAD, line 24: Let's pass the version from here, not the entire flags

yarn.lock
```diff
@@ -2,324 +2,341 @@
...
-    "@babel/generator" "^7.16.7"
-    "@babel/helper-compilation-targets" "^7.16.7"
-    "@babel/helper-module-transforms" "^7.16.7"
-    "@babel/helpers" "^7.16.7"
-    "@babel/parser" "^7.16.7"
-    "@babel/template" "^7.16.7"
-    "@babel/traverse" "^7.16.7"
-    "@babel/types" "^7.16.7"
+  version "7.15.8"
+  resolved "https://registry.npmjs.org/@babel/core/-/core-7.15.8.tgz"
```
RodrigoAD: Make sure to always use `yarn` instead of `npm` for installing and executing

packages-ts/gauntlet-terra-contracts/src/lib/contracts.ts
```diff
@@ -30,36 +31,27 @@ export type Contract = {
...
 }
 
-// TODO: Pull it from a Github versioned release artifact
-export const getContractCode = (contractId: CONTRACT_LIST): string => {
-  // Possible paths depending on how/where gauntlet is being executed
-  const possibleContractPaths = [
-    path.join(__dirname, '../..', './artifacts/bin'),
-    path.join(process.cwd(), './artifacts/bin'),
-    path.join(process.cwd(), './packages-ts/gauntlet-terra-contracts/artifacts/bin'),
-  ]
```
krebernisak, line 55: We should still leave this option open while in the development of the next contract/gauntlet version, right @RodrigoAD? We could also use an explicit version like `next` or `local` to configure this mode.

krebernisak: Let's also make sure we use the same artifact loading workflow on Solana.

RodrigoAD: Yeah good point. Let's leave the option with `local` mode.

We could do this exact change next on Solana

PR #55, opened 2022-01-26
Testnet soak cleanup

connorwstein:


jmank88 (dismissed):

jmank88 (dismissed):

pkg/terra/client/client.go
```diff
@@ -59,8 +59,13 @@ type Writer interface {
...
 const (
-	DefaultTimeout            = 10 * time.Second
-	DefaultGasLimitMultiplier = 1.5
+	// DefaultTimeout is the default Terra client timeout.
+	// Note that while the terra node is processing a heavy block,
+	// requests can be delayed significantly (https://github.com/tendermint/tendermint/issues/6899),
+	// however there's nothing we can do but wait until the block is processed.
+	// So we set a fairly high timeout here.
+	DefaultTimeout            = 30 * time.Second
+	DefaultGasLimitMultiplier = 1.2
```
krebernisak, line 68: The test is failing because this value is too low now "gasWanted: 62049, gasUsed: 62118: out".

pkg/terra/config.go
```diff
@@ -18,7 +18,12 @@ var defaultConfigSet = configSet{
 	BlocksUntilTxTimeout:  10,
 	ConfirmPollPeriod:     time.Second,
 	FallbackGasPriceULuna: sdk.MustNewDecFromStr("0.015"),
-	GasLimitMultiplier:    1.5,
+	// This is high since we simulate before signing the transaction.
+	// There's a chicken and egg problem: need to sign to simulate accurately
+	// but you need to specify a gas limit when signing.
+	// TODO: Determine how much gas a signature adds and then
+	// add that directly so we can be more accurate.
+	GasLimitMultiplier: 1.5,
```
krebernisak, line 26: Should we use `DefaultGasLimitMultiplier` here to have a single source of truth?

pkg/terra/config.go
```diff
@@ -18,7 +18,12 @@ var defaultConfigSet = configSet{
 	BlocksUntilTxTimeout:  10,
 	ConfirmPollPeriod:     time.Second,
 	FallbackGasPriceULuna: sdk.MustNewDecFromStr("0.015"),
-	GasLimitMultiplier:    1.5,
+	// This is high since we simulate before signing the transaction.
```
krebernisak, line 21: hmm, should we also have `DefaultTimeout` configurable?

krebernisak: We can def defer this for later...

jmank88 (dismissed):

pkg/terra/config.go
connorwstein: yeah it should be, can be done in a follow up

PR #56, opened 2022-01-26
log malformed jsonrpc responses

jmank88:


connorwstein (commented):
lgtm - can we test it against the bison trails node without auth?

connorwstein (approved):

PR #69, opened 2022-01-31
event attribute parsing helper

### LINKED ISSUE #11: Investigate if there is a better way to parse an array of structs to an struct
Consider this snippet in `pkg/terra/contract_config_tracker.go`:

```go
	for _, event := range res.Txs[0].TxResult.Events {
		if event.Type == "wasm-set_config" {
			output := types.ContractConfig{}
			// TODO: is there a better way to parse an array of structs to an struct
			for _, attr := range event.Attributes {
				key, value := string(attr.Key), string(attr.Value)
				switch key {
```

jmank88:
Closes #11 - this appears to be efficient already, so I just separated out a helper func to reduce nesting and improve readability.


krebernisak (approved):
LGTM👌

PR #70, opened 2022-01-31
Wait for config

connorwstein:


pkg/terra/contract_cache.go
```diff
@@ -94,12 +95,23 @@ func (cc *ContractCache) updateConfig(ctx context.Context) error {
...
+	var same, firstTimeConfig bool
 	cc.configMu.Lock()
-	same := cc.configBlock == changedInBlock && cc.config.ConfigDigest == configDigest
-	if same {
-		cc.configTS = now // refresh TTL
+	{
+		if cc.configBlock == 0 && changedInBlock != 0 {
+			cc.lggr.Infof("detected first time configuration")
+			firstTimeConfig = true
+		}
```
jmank88, line 104: What about checking for `configTS.IsZero()` instead? And logging down below, outside of the lock?
```suggestion
		firstTimeConfig = configTS.IsZero()
```

connorwstein: Pulled the signal check out into the poller loop as I think it's more readable there and just added a bool to ensure we only signal once

connorwstein:
closing in favor of simpler https://github.com/smartcontractkit/chainlink-terra/pull/259

PR #72, opened 2022-02-01
update commands to await on getContract

akhilchainani:
fix for the failing commands due to a mismatch type

### DIFF
--- a/packages-ts/gauntlet-terra-contracts/src/lib/contracts.ts
+++ b/packages-ts/gauntlet-terra-contracts/src/lib/contracts.ts
@@ -31,21 +31,6 @@ export type Contract = {
 
 export type Contracts = Record<CONTRACT_LIST, Contract>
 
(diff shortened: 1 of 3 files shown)


packages-ts/gauntlet-terra-contracts/src/lib/contracts.ts
```diff
@@ -31,19 +31,19 @@ export type Contract = {
 
 export type Contracts = Record<CONTRACT_LIST, Contract>
 
-export const loadContracts = (version): Contracts => {
-  return Object.values(CONTRACT_LIST).reduce((agg, id) => {
+export const loadContracts = async (version): Promise<Contracts> => {
```
RodrigoAD, line 34: We can remove this function, don't think we use it anymore

akhilchainani: Good catch! just fixed this

RodrigoAD (approved):

PR #73, opened 2022-02-01
Fix double b64 encoding signers

jmank88:


archseer:
Thanks! I'll also cherry-pick this into my branch

archseer (approved):

PR #74, opened 2022-02-01
Get a gauntlet e2e test in ci

tateexon:
Starting with just the upload command


tests/e2e/smoke/gauntlet_test.go
```diff
@@ -0,0 +1,104 @@
...
+			g.NetworkConfig = common.GetDefaultGauntletConfig(terraNodeUrl)
+			g.WriteNetworkConfigMap()
+		})
+	})
+
+	Describe("Run Gauntlet Commands", func() {
+		It("should upload the contracts", func() {
+			_, err = g.ExecCommandWithRetries([]string{
+				"upload",
+				g.Flag("version", "local"),
```
RodrigoAD, line 87: We should probably specify the network flag here. 
Actually, I think the Action failed, but gave a false positive cause this incorrect `process.exit` ? https://github.com/smartcontractkit/chainlink-terra/runs/5029805743?check_suite_focus=true#step:5:19

tateexon: The network is always applied: https://github.com/smartcontractkit/integrations-framework/blob/main/gauntlet/gauntlet.go#L69
with a unique one generated for each test when calling NewGauntlet: https://github.com/smartcontractkit/integrations-framework/blob/main/gauntlet/gauntlet.go#L46

As for that particular failure you linked to, my tests have nothing to do with that run although my gauntlet test did have a false positive and didn't run on pull request. I will make it run on pull request and see if I can figure out my own false positive. I will also add that false positive you pointed out to my todo list!

tateexon:
The test is now running on pull requests and passing now. It is ready for another look @RodrigoAD @skudasov

tests/e2e/smoke/gauntlet_test.go
RodrigoAD: You're right, I looked at the wrong test 😅

RodrigoAD (approved):
This is starting to look really good :)

PR #75, opened 2022-02-02
single global EncodingConfig

jmank88:


pkg/terra/client/client.go
```diff
@@ -444,3 +442,8 @@ func (c *Client) Balance(addr sdk.AccAddress, denom string) (*sdk.Coin, error) {
 	}
 	return b.Balance, nil
 }
+
+//TODO how much of app.MakeEncodingConfig to reproduce?
```
jmank88, line 446: `app.MakeEncodingConfig()` is documented for testing only and calls global `Register*` funcs, so we shouldn't use it in production and we can't call it more than once anyways. `params.MakeEncodingConfig` swaps right in, but I don't know if we depend on the global register calls too (and would need to reproduce calling those once from an init).

jmank88: Test failures suggest we need those, so added to a `func init()`.

pkg/terra/client/client.go
```diff
@@ -12,27 +12,42 @@ import (
...
+	encodingConfig := params.MakeEncodingConfig()
+	std.RegisterLegacyAminoCodec(encodingConfig.Amino)
+	std.RegisterInterfaces(encodingConfig.InterfaceRegistry)
+	app.ModuleBasics.RegisterLegacyAminoCodec(encodingConfig.Amino)
+	app.ModuleBasics.RegisterInterfaces(encodingConfig.InterfaceRegistry)
+
+	// authz module use this codec to get signbytes.
+	// authz MsgExec can execute all message types,
+	// so legacy.Cdc need to register all amino messages to get proper signature
+	app.ModuleBasics.RegisterLegacyAminoCodec(legacy.Cdc)
```
jmank88, line 48: https://github.com/terra-money/core/blob/main/app/encoding.go#L12-L29

pkg/terra/client/client.go
```diff
@@ -12,27 +12,43 @@ import (
...
 	"github.com/smartcontractkit/terra.go/key"
+	"github.com/smartcontractkit/terra.go/msg"
 	"github.com/smartcontractkit/terra.go/tx"
 )
 
+var encodingConfig = params.MakeEncodingConfig()
+
+func init() {
+	// Based on app.MakeEncodingConfig
+	std.RegisterLegacyAminoCodec(encodingConfig.Amino)
```
connorwstein, line 41: let's add a comment that these functions can panic when called more than once

connorwstein (approved):

pkg/terra/client/client.go
jmank88: :heavy_check_mark:

connorwstein (approved):

PR #80, opened 2022-02-04
Add known good versions for node, go, rust to asdf file and use nix for ci e2e tests

tateexon:
Had some issues with people running incompatible node versions so this should help them be able to build and run tests


.github/workflows/e2e.yml
```diff
@@ -40,6 +40,10 @@ jobs:
         uses: actions/setup-go@v1
         with:
           go-version: 1.17
+      - name: Setup Node Version
+        uses: actions/setup-node@v2
+        with:
+          node-version-file: '.nvmrc'
```
archseer, line 46: Given that the other actions use Nix, could e2e do the same?

krebernisak: Yes, please!

I'd suggest all our automation (CI workflows) use Nix, and we additionally standardize on [asdf](https://asdf-vm.com) for local dev environments as it's powerful enough to support our polyglot use cases. It's also loved by many Chainlink engineers.

@tateexon why don't we turn this effort into:

1. Propagate Nix usage in CI workflows
2. Configure & document `asdf` local environments

Ideally, any progress we also consistently apply to CL Solana.

Here is a nice article on [Better project environment management with direnv and asdf](https://blog.mikecordell.com/2021/12/18/better-project-environment-management-with-direnv-and-asdf/).

krebernisak (changes requested):

archseer:
You might get better luck with replacing `pkg.mkShell { .. }` with something similar to

```nix
(pkgs.buildFHSUserEnv {
  name = "chainlink-shell";
  targetPkgs = pkgs: with pkgs; [
    glibc
    stdenv.cc.cc
    ... deps here
  ];
}).env
```

It's weird that you're getting this though, is the binary built outside of the nix environment? Or is the binary executed outside of it? It seems to be using a newer glibc when building, then it defaults to the old one available on the host when executing.

archseer:
Okay weird, looks like it worked for you without `LD_LIBRARY_PATH` specified :O

RodrigoAD (approved):

krebernisak (approved):
Nice work! Please check/replicate the same for our Solana repo 🙌

archseer:
> , not sure if it was placed originally in here on purpose or a copy paste from another repo.

From what I remember certain compiled binaries (Go) don't work for me locally without that line.

PR #85, opened 2022-02-07
Feature/27989 Ledger tx signature

sdrug:


packages-ts/gauntlet-terra/src/commands/ledgerKey.ts
```diff
@@ -0,0 +1,60 @@
...
+import LedgerTerraConnector, {ERROR_CODE} from '@terra-money/ledger-terra-js'
+import TransportNodeHid from "@ledgerhq/hw-transport-node-hid"
+import { logger } from '@chainlink/gauntlet-core/dist/utils'
+
+
+export class LedgerKey extends Key {
+    private path: Array<number>
+    private ledgerConnector: LedgerTerraConnector
+
+    constructor(path: Array<number>) {
```
RodrigoAD, line 11: Can you make the constructor `private` ? We should force to use the `create` fn to initialize the class

packages-ts/gauntlet-terra/src/commands/ledgerKey.ts
```diff
@@ -0,0 +1,66 @@
...
+        } catch (e) {
+            logger.error(`LedgerKey sign failed: ${e.message}. Is Ledger unlocked and Terra app open?`)
+            throw e
+        }
+    }
+
+    private async connectToLedger(){
+        const transport = await TransportNodeHid.create()
+        const ledgerConnector = new LedgerTerraConnector(transport)
+        await ledgerConnector.initialize()
```
RodrigoAD, line 59: Seems that this is already returning if the connection succeeded. (https://github.com/terra-money/ledger-terra-js/blob/master/src/app.ts#L102). We could check it here instead of [here](https://github.com/smartcontractkit/chainlink-terra/pull/85/files#diff-d9d0318909e07b911c8d2549f846f8fe746e654351ac85c1aaca9f9b986f7484R18)

packages-ts/gauntlet-terra/src/commands/ledgerKey.ts
```diff
@@ -0,0 +1,66 @@
...
+    private static pathStringToArray(path: string): Array<number> {
+        return path.split(',').map(item => parseInt(item))
+    }
+
+    public async sign(payload: Buffer): Promise<Buffer> {
+        try { 
+            const {ledgerConnector} = await this.connectToLedger()
+
+            const response = await ledgerConnector.sign(this.path, payload)
+            if (response.return_code !== ERROR_CODE.NoError) {
```
RodrigoAD, line 45: We are repeating the error check several times. Try to find a better way to handle these. Maybe wrapping the ledger functions, and check the response in the wrapper

packages-ts/gauntlet-terra/src/commands/ledgerKey.ts
```diff
@@ -0,0 +1,66 @@
...
+
+    public static async create(path: string): Promise<LedgerKey> {
+        const pathArr = this.pathStringToArray(path)
+        const ledgerKey = new LedgerKey(pathArr)
+        await ledgerKey.initialize()
+
+        return ledgerKey
+    }
+
+    private static pathStringToArray(path: string): Array<number> {
```
RodrigoAD, line 36: Users should be able to enter the path on the standard format (`44'/60'/0'/0'/0`). Can we support that?

packages-ts/gauntlet-terra/src/commands/ledgerKey.ts
```diff
@@ -0,0 +1,66 @@
...
+    public async initialize() {
+        const {ledgerConnector, terminateConnection} = await this.connectToLedger()
+
+        const response = await ledgerConnector.getPublicKey(this.path)
+        if (response.return_code !== ERROR_CODE.NoError) {
+            logger.error(`LedgerKey init failed: ${response.error_message}. Is Ledger unlocked and Terra app open?`)
+            throw new Error(response.error_message)
+        }
+
+        this.publicKey = new SimplePublicKey(Buffer.from(response.compressed_pk.data).toString('base64'))
```
RodrigoAD, line 24: This doesn't seem to be used, no?

packages-ts/gauntlet-terra/src/commands/ledgerKey.ts
sdrug: Hmm, I didn't quite understand here what you mean. We could check what?

packages-ts/gauntlet-terra/src/commands/ledgerKey.ts
sdrug: done

packages-ts/gauntlet-terra/src/commands/ledgerKey.ts
sdrug: done

packages-ts/gauntlet-terra/src/commands/ledgerKey.ts
sdrug: done

packages-ts/gauntlet-terra/src/commands/ledgerKey.ts
RodrigoAD: `ledgerConnector.initialize()` returns a response if the connection failed. We should check the response an throw in case of error

packages-ts/gauntlet-terra/src/commands/ledgerKey.ts
```diff
@@ -0,0 +1,118 @@
...
+
+    public static async create(path: string): Promise<LedgerKey> {
+        const pathArr = this.pathStringToArray(path)
+        const ledgerKey = new LedgerKey(pathArr)
+        await ledgerKey.initialize()
+
+        return ledgerKey
+    }
+
+    public async createSignature(signDoc: SignDoc): Promise<SignatureV2> {
```
RodrigoAD, line 35: Is there any difference between this and the base `Key` `createSignature` function?

sdrug: yes, the difference is in data and data format supplied to the sign function: Ledger requires amino encoded

packages-ts/gauntlet-terra/src/commands/ledgerKey.ts
```diff
@@ -0,0 +1,118 @@
...
+            return
+
+        let errorMessage: string
+        if (isLocked) {
+            errorMessage = 'Is Ledger unlocked and Terra app open?'
+        } else {
+            errorMessage = ledgerError
+        }
+
+        throw new Error(errorMessage)
```
RodrigoAD, line 116:
```suggestion
        if (isLocked) {
            throw new Error('Is Ledger unlocked and Terra app open?')
        }
        throw new Error(ledgerError)
```

packages-ts/gauntlet-terra/src/commands/ledgerKey.ts
```diff
@@ -0,0 +1,118 @@
...
+      }
+
+    public async sign(payload: Buffer): Promise<Buffer> {
+        const {ledgerConnector, terminateConnection} = await this.connectToLedger()
+        try { 
+            logger.info('Approve tx on your Ledger device.')
+            const response = await ledgerConnector.sign(this.path, payload)
+            this.checkForErrors(response)
+   
+            const signature = signatureImport(Buffer.from(response.signature as any))
```
RodrigoAD, line 74:
```suggestion
            const signature = signatureImport(Buffer.from(response.signature.data))
```

RodrigoAD (dismissed):
Really good work @sdrug. Once the CI is green we can merge :)

packages-ts/gauntlet-terra/src/lib/constants.ts
```diff
@@ -1 +1,2 @@
 export const ADDRESS_ZERO = 'terra000000000000000000000000000000000000000'
+export const LEDGER_ULUNA_PATH = "44'/330'/0'/0'/0"
```
krebernisak, line 2: First, this should ref LUNA not ULUNA, but also it's not following the [BIP44 spec](https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki)., which defines the derivation path as:

```
m / purpose' / coin_type' / account' / change / address_index
```

Notice the extra `'` in your path declaration.

Additionally, the BIP44 is not Ledger specific in any way and can be used with a local private key also if we add HD capabilities to Gauntlet.

```suggestion
export const BIP44_LUNA_PATH = "44'/330'/0'/0/0"
```

krebernisak (changes requested):

packages-ts/gauntlet-terra/src/commands/ledgerKey.ts
```diff
@@ -0,0 +1,78 @@
...
+      ledgerConnector,
+      terminateConnection: transport.close.bind(transport),
+    }
+  }
+
+  private static pathStringToArray(path: string): Array<number> {
+    return path.split("'/").map((item) => parseInt(item))
+  }
+
+  private checkForErrors(response: CommonResponse) {
```
krebernisak, line 66:
```suggestion
  private assertNoErrors(response: CommonResponse) {
```

or

```suggestion
  private throwOnError(response: CommonResponse) {
```

packages-ts/gauntlet-terra/src/commands/internal/terra.ts
```diff
@@ -60,6 +62,9 @@ export default abstract class TerraCommand extends WriteCommand<TransactionRespo
 
     const tx = await this.wallet.createAndSignTx({
       msgs: [msg],
+      ...(this.wallet.key instanceof LedgerKey && {
```
krebernisak, line 65: Consider defining this in the constructor once, and use it multiple times.

packages-ts/gauntlet-terra/src/lib/constants.ts
krebernisak: GH Issue (on Solana):

- https://github.com/smartcontractkit/chainlink-solana/issues/193

packages-ts/gauntlet-terra/src/commands/ledgerKey.ts
```diff
@@ -51,19 +53,23 @@ export class LedgerKey extends Key {
...
 
     return {
       ledgerConnector,
       terminateConnection: transport.close.bind(transport),
     }
   }
 
-  private static pathStringToArray(path: string): Array<number> {
-    return path.split("'/").map((item) => parseInt(item))
+  private static bip44PathtoArray(path: string): Array<number> {
```
krebernisak, line 64: This is a nice util function that we should move to `gauntlet-core` at one point.

krebernisak (approved):

PR #95, opened 2022-02-10
Add juelsperuluna, aggregatorroundid, failed/succeded tx counts

topliceanu:
New env var required:
```
TERRA_FCD_URL="https://fcd.terra.dev/" 
```


pkg/monitoring/source_envelope.go
```diff
@@ -122,33 +116,35 @@ func (s *terraSource) Fetch(ctx context.Context) (interface{}, error) {
...
-		balance, err := strconv.ParseUint(balanceRes.Balance, 10, 64)
-		if err != nil {
-			envelopeErr = multierr.Combine(envelopeErr, fmt.Errorf("failed to parse uint64 balance from '%s': %w", balanceRes.Balance, err))
+		balance, success := new(big.Int).SetString(balanceRes.Balance, 10)
+		if !success {
+			envelopeErr = multierr.Combine(fmt.Errorf("failed to parse link balance from '%s'", balanceRes.Balance))
 			return
 		}
-		envelope.LinkBalance = balance
+		envelope.LinkBalance = uint64(balance.Int64())
```
jmank88, line 124: Why the switch from parsing directly as uint64? Should this conversion be checked first with `IsInt64()`, since it's undefined otherwise? (And/or could `Uint64()/IsUint64()` be used?)

pkg/monitoring/source_txresults.go
```diff
@@ -0,0 +1,117 @@
...
+				maxTxID = tx.ID
+			}
+		}
+		t.latestTxID = maxTxID
+	}()
+	// Count failed and succeeded recent transactions
+	output := relayMonitoring.TxResults{}
+	for _, tx := range recentTxs {
+		// See https://github.com/terra-money/core/blob/main/x/wasm/types/errors.go
+		if tx.Code > 2 && tx.CodeSpace == "wasm" {
```
jmank88, line 110: I think we have seen `Codespace == "sdk"` failures as well, but maybe only as broadcast rejections, not on-chain failures... :thinking:

topliceanu: Maybe a better solution is to count the non-errors, ei. txs that don't have a `Code`. Wdyt?

jmank88: Yeah that seems safer, and it looks like `Code==1` is a special internal value too:
https://docs.cosmos.network/master/building-modules/errors.html

pkg/monitoring/source_envelope.go
topliceanu: I had a convo with @kaleofduty and these balances - as well as the observations - will more than fit in a float64.

topliceanu: So going forward, we're going to read big.Ints from the chain; convert them to int64 - without the IsInt64 check - and then from there to a float64 ready for export to prometheus.

pkg/monitoring/source_txresults.go
topliceanu: done

pkg/monitoring/source_txresults.go
```diff
@@ -0,0 +1,126 @@
...
+		return nil, fmt.Errorf("unable to decode transactions from response: %w", err)
+	}
+	// Filter recent transactions
+	recentTxs := []fcdTx{}
+	func() {
+		t.latestTxIDMu.Lock()
+		defer t.latestTxIDMu.Unlock()
+		maxTxID := t.latestTxID
+		for _, tx := range txsResponse.Txs {
+			if tx.ID > t.latestTxID {
```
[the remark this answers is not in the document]
topliceanu: It doesn't need to be sorted. Note that I'm doing two things here, 1. remove old transactions that are before `t.latestTxID` and 2. compute latest tx id - `maxTxID` and update `t.latestTxID`. 
I agree this code is not very clear because it merges these two goals. I did this to grab the lock only once.

PR #97, opened 2022-02-10
Support negatives in min/max

connorwstein:


pkg/terra/utils.go
```diff
@@ -89,10 +97,40 @@ func ContractConfigToOCRConfig(in []byte) ([]byte, error) {
...
+		return nil, err
+	}
+	return bytes.Join([][]byte{version, minBytes, maxBytes}, []byte{}), nil
+}
+
+func parse128BitSignedInt(s []byte) (*big.Int, error) {
+	if len(s) != byteWidth128 {
+		return nil, fmt.Errorf("invalid int length: expected %d got %d", byteWidth128, len(s))
+	}
+	val := (&big.Int{}).SetBytes(s)
```
jmank88, line 124:
```suggestion
	val := new(big.Int).SetBytes(s)
```

pkg/terra/utils_test.go
```diff
@@ -188,34 +192,71 @@ func TestRawMessageStringIntToInt(t *testing.T) {
...
 		})
 	}
 }
+
+func TestParse128Bit(t *testing.T) {
+	b, err := hex.DecodeString("80000000000000000000000000000000")
+	require.NoError(t, err)
+	i, err := parse128BitSignedInt(b)
+	require.NoError(t, err)
+	fmt.Println(i)
```
jmank88, line 260:
```suggestion
```

pkg/terra/utils.go
```diff
@@ -89,10 +97,40 @@ func ContractConfigToOCRConfig(in []byte) ([]byte, error) {
...
+	// 2**127 - 1
+	maxPositive := big.NewInt(0).Sub(big.NewInt(0).Lsh(big.NewInt(1), bitWidth128-1), big.NewInt(1))
```
jmank88, line 126: Could move to global variable so it's not recomputed.

pkg/terra/utils.go
```diff
@@ -89,10 +97,40 @@ func ContractConfigToOCRConfig(in []byte) ([]byte, error) {
...
-	return bytes.Join([][]byte{version, padding, min128, padding, max128}, []byte{}), nil
+	min, err := parse128BitSignedInt(in[1 : byteWidth128+1])
+	if err != nil {
+		return nil, err
+	}
+	max, err := parse128BitSignedInt(in[1+byteWidth128:])
+	if err != nil {
+		return nil, err
+	}
+	minBytes, err := median.ToBytes(min)
```
krebernisak, line 109: Hmm, is there an opportunity for us to use [OnchainConfig.Encode()](https://github.com/smartcontractkit/libocr/blob/5b957c8342763a061fcbd6806ef98351f3cc7f93/offchainreporting2/reportingplugin/median/median.go#L54) directly?

pkg/terra/utils.go
```diff
@@ -89,10 +97,40 @@ func ContractConfigToOCRConfig(in []byte) ([]byte, error) {
...
+		return nil, err
+	}
+	maxBytes, err := median.ToBytes(max)
+	if err != nil {
+		return nil, err
+	}
+	return bytes.Join([][]byte{version, minBytes, maxBytes}, []byte{}), nil
+}
+
+func parse128BitSignedInt(s []byte) (*big.Int, error) {
```
krebernisak, line 120: On Solana, I think we use `gagliardetto/binary` [Int128](https://github.com/gagliardetto/binary/blob/master/u128.go#L195) type. Could we also reuse that here?

jmank88 (dismissed):

krebernisak (approved):
Nice work! 👍 

CI is all green, but there are some changes to the on-chain config encoding, upstream by @archseer, so merging this in first.

PR #111, opened 2022-02-14
Multisig wrapper command

RodrigoAD:


packages-ts/gauntlet-terra-cw20-multisig/package.json
```diff
@@ -0,0 +1,32 @@
+{
+  "name": "@chainlink/gauntlet-terra-cw20-multisig",
```
krebernisak, line 2: Let' s rename this package and have it support the cw-plus contracts we need:

- cw3-flex-multisig
- cw4-group
- cw20-base

```suggestion
  "name": "@chainlink/gauntlet-terra-cw-plus",
```

packages-ts/gauntlet-terra/src/commands/internal/terra.ts
```diff
@@ -63,6 +64,25 @@ export default abstract class TerraCommand extends WriteCommand<TransactionRespo
...
+        }),
+      })
+
+      const res = await this.provider.tx.broadcast(tx)
+
+      logger.debug(res)
+      return this.wrapResponse(res)
+    } catch (e) {
+      const details = e.response.data
+      throw new Error(details.message)
```
krebernisak, line 82: What's the purpose of this catch-throw?

Also if there is a reason to leave it in, we should use safe navigation:
```suggestion
      throw new Error(e.response.?data.?message)
```

packages-ts/gauntlet-terra/src/commands/internal/terra.ts
```diff
@@ -63,6 +64,25 @@ export default abstract class TerraCommand extends WriteCommand<TransactionRespo
...
+      const tx = await this.wallet.createAndSignTx({
+        msgs: messages,
+        ...(this.wallet.key instanceof LedgerKey && {
+          signMode: SignMode.SIGN_MODE_LEGACY_AMINO_JSON,
+        }),
+      })
+
+      const res = await this.provider.tx.broadcast(tx)
+
+      logger.debug(res)
```
krebernisak, line 78: This debug log looks out of place, do we want it? Will this only appear with `--debug` flag?

packages-ts/gauntlet-terra-contracts/src/commands/abstract/inspectionWrapper.ts
```diff
@@ -47,6 +47,10 @@ export const instructionToInspectCommand = <CommandInput, Expected>(
       super(flags, args)
     }
 
+    makeRawTransaction = () => {
+      throw new Error('Inspection command does not involve any transaction')
```
krebernisak, line 51: Yeah, read functions should be of a different type.

packages-ts/gauntlet-terra-cw20-multisig/src/commands/multisig.ts
```diff
@@ -0,0 +1,250 @@
...
+
+    constructor(flags, args) {
+      super(flags, args)
+
+      this.command = new command(flags, args)
+
+      if (!AccAddress.validate(process.env.MULTISIG_ADDRESS)) throw new Error(`Invalid Multisig wallet address`)
+      if (!AccAddress.validate(process.env.MULTISIG_GROUP)) throw new Error(`Invalid Multisig group address`)
+      this.multisig = process.env.MULTISIG_ADDRESS as AccAddress
+      this.multisigGroup = process.env.MULTISIG_GROUP as AccAddress
```
krebernisak, line 66: This should really be read from the input, and not directly from env vars here. Env vars should be completely abstracted from the command. The command just gets some kind of input (flags) that can be loaded from anywhere (CLI flags/args, env, file ...)

packages-ts/gauntlet-terra-cw20-multisig/src/commands/multisig.ts
```diff
@@ -0,0 +1,250 @@
...
+      const proposeInput = {
+        propose: {
+          description: command.id,
+          msgs: [this.toWasmMsg(message)],
+          title: command.id,
+          // TODO: Set expiration time
+          // latest: { never: {} }
+        },
+      }
+      return new MsgExecuteContract(signer, this.multisig, proposeInput)
```
krebernisak, line 118: Where do we define the wasm function to call? e.g. `begin_proposal`?

packages-ts/gauntlet-terra-cw20-multisig/src/commands/multisig.ts
```diff
@@ -0,0 +1,250 @@
...
+      const approvalInput = {
+        vote: {
+          vote: Vote.YES,
+          proposal_id: proposalId,
+        },
+      }
+      return new MsgExecuteContract(signer, this.multisig, approvalInput)
+    }
+
+    executeExecution: ProposalAction = async (signer, proposalId) => {
```
krebernisak, line 132: Execute execution? :D

krebernisak: None of these are "executing" anything, but just building payloads. Consider renaming.

packages-ts/gauntlet-terra/src/commands/internal/terra.ts
RodrigoAD: The default `e.message` gives almost no information. Adding the safe navigation

packages-ts/gauntlet-terra/src/commands/internal/terra.ts
RodrigoAD: Yeah we don't really need it, removing it

packages-ts/gauntlet-terra-cw20-multisig/src/commands/multisig.ts
```diff
@@ -0,0 +1,275 @@
...
+      const approvalsLeft = state.threshold - state.approvers.length
+      const messages = {
+        [Action.APPROVE]: `The proposal needs ${approvalsLeft} more approvals. Run the same command with the flag --proposal=${proposalId}`,
+        [Action.EXECUTE]: `The proposal reached the threshold and can be executed. Run the same command with the flag --proposal=${proposalId}`,
+        [Action.NONE]: `The proposal has been executed. No more actions needed`,
+      }
+      logger.line()
+      logger.info(`${messages[state.nextAction]}`)
+      logger.line()
+    }
```
krebernisak, line 198: This is something we used to call an inspection command.

Here it's just a function, on a different command class. 
The Multisig command, that groups all Multisig abstract commands?

How do I call this inspection command? How do I compose it in a workflow?
Is the Multisig command a workflow or is it something else? What exactly?

I have many questions :)

...

The first priority is to have it function in any way possible, and while this looks ok IMO we need to go back to the Gauntlet v4 drawing board and start typing and composing these commands better.

packages-ts/gauntlet-terra-cw20-multisig/src/commands/multisig.ts
```diff
@@ -0,0 +1,275 @@
...
+      logger.info(`Proposal State:
+        - Total Owners: ${state.owners.length}
+        - Owners List: ${state.owners}
+
+        - Threshold: ${state.threshold}
+        - Total Approvers: ${state.approvers.length}
+        - Approvers List: ${state.approvers}
+
+        - Next Action: ${state.nextAction.toUpperCase()}
+      `)
```
krebernisak, line 219: This is another inspection command baked in ad-hoc.

packages-ts/gauntlet-terra-cw20-multisig/src/commands/multisig.ts
RodrigoAD: Hehe yes, I'll rename it

packages-ts/gauntlet-terra-cw20-multisig/src/commands/multisig.ts
yosriady: Regarding loading environment variables, how about something like this:

```
const environment: Environment = load(.env filepath)

...

executeCLI(environment, myCommands)
```

Then, in each command:

```
this.multisig = this.env.MULTISIG_ADDRESS
```

where `this.env` contains the environment variables, available in each `Command` instance.

packages-ts/gauntlet-terra-cw20-multisig/src/commands/multisig.ts
RodrigoAD: Not sure is this is what you mean, but here we decide which operation to call: https://github.com/smartcontractkit/chainlink-terra/pull/111/files#diff-c7c2eafb6b93bd0e456c2fd133626df8051e82da80db2ad3695f45fe6cbd72bbR88
That's based on `state.nextAction`

packages-ts/gauntlet-terra-cw20-multisig/src/commands/multisig.ts
RodrigoAD: I see your point. This goes a bit beyond from this PR. Opening an issue for it: #133

packages-ts/gauntlet-terra-cw20-multisig/src/commands/multisig.ts
RodrigoAD: #133

packages-ts/gauntlet-terra-cw20-multisig/src/commands/multisig.ts
RodrigoAD: Opening an issue here, thought could be a simpler fix, but Gauntlet is not really prepared to inject env vars anywhere.
https://github.com/smartcontractkit/chainlink-terra/issues/134

PR #113, opened 2022-02-15
pkg/terra: add MsgEnqueuer.GetMsgs

jmank88:
Supports https://github.com/smartcontractkit/chainlink/pull/6003


pkg/terra/relay.go
```diff
@@ -25,7 +35,13 @@ type Logger interface {
 }
 
 type MsgEnqueuer interface {
-	Enqueue(contractID string, msg []byte) (int64, error)
+	// Enqueue enqueues msg for broadcast and returns its id.
```
connorwstein, line 38: is there no way to avoid expanding this interface? I was hoping to keep it to actually just the Enqueue method

jmank88: Yeah, could use a separate or extended interface.

jmank88: Updated with separate TxManager, and dropped Start/Close methods.

pkg/terra/client/test_helpers.go
```diff
@@ -75,49 +78,59 @@ func SetupLocalTerraNode(t *testing.T, chainID string) ([]Account, string) {
...
 	for i := 0; i < 10; i++ {
-		time.Sleep(1 * time.Second)
-		out, err = exec.Command("curl", "http://127.0.0.1:26657/abci_info").Output()
-		require.NoError(t, err)
+		time.Sleep(time.Second)
+		out, err = exec.Command("curl", tendermintURL+"/abci_info").Output() //nolint:gosec
+		if err != nil {
+			t.Logf("API server not ready yet (attempt %d): %v\n", i+1, err)
+			continue
+		}
```
jmank88, line 110: This require was a bit too strict, since calling 'too soon' returns an exit code 7.

pkg/terra/client/test_helpers.go
```diff
@@ -75,49 +79,73 @@ func SetupLocalTerraNode(t *testing.T, chainID string) ([]Account, string) {
...
+	cmd.Stderr = &stdErr
 	require.NoError(t, cmd.Start())
 	t.Cleanup(func() {
-		require.NoError(t, cmd.Process.Kill())
+		assert.NoError(t, cmd.Process.Kill())
+		if err2 := cmd.Wait(); assert.Error(t, err2) {
+			if !assert.Contains(t, err2.Error(), "signal: killed", cmd.ProcessState.String()) {
+				t.Log("terrad stderr:", stdErr.String())
+			}
+		}
```
jmank88, line 114: This was the key to diagnosing the core CI failures.

connorwstein: curious what the error string was?

connorwstein (approved):

pkg/terra/client/test_helpers.go
jmank88: It would immediately exiting complaining about one of the ports, and dump a command help menu, but that didn't trigger a normal error.

PR #114, opened 2022-02-15
Fix monitoring on-chain config

krebernisak:


topliceanu:
This is fixed in this PR https://github.com/smartcontractkit/chainlink-terra/pull/95

krebernisak:
@topliceanu yup, saw it just that `onchain_config` handling change so I wanted to propagate it to monitoring, as there were compile errors CI didn't catch. So I suggest we merge this in first, and then you rebase your PR on top.

PR #116, opened 2022-02-15
Multisig wrapper

RodrigoAD:


packages-ts/gauntlet-terra-contracts/src/commands/abstract/polymorphic.ts
```diff
@@ -0,0 +1,64 @@
...
+import AbstractCommand from '.'
+import { makeAbstractCommand } from '.'
+
+class EmptyCommand extends Command {
+  constructor() {
+    super({ help: false }, [])
+  }
+}
+
+type ICommandConstructor = (flags: any, args: string[]) => void
```
RodrigoAD, line 15: Why this "polymorphic" command? Why do we need it? We are already wrapping the commands into a multisig command

packages-ts/gauntlet-terra-contracts/src/commands/abstract/polymorphic.ts
```diff
@@ -0,0 +1,64 @@
...
+            this.command = new commandType(flags, args)
+          }
+        }
+
+        multisigOp = () => {
+          return op
+        }
+      }
+
+      // This is a temporary workaround for a bug in the type specification for findPolymorphic in @gauntlet-core.
```
RodrigoAD, line 53: Polymorphic is a legacy command that gauntlet-core had to support for old Gauntlet EVM, but we won't be using it in the future

packages-ts/gauntlet-terra-contracts/src/commands/contracts/multisig/multisig.ts
```diff
@@ -0,0 +1,67 @@
+// multisig.ts
+//
+// For now, propose, vote, and execute functionality are all combined into one CONTRACT:COMMAND::multisig meta-command
+//  This is parallel to how things are implemented in Solana.  The execute happens automatically as soon as the last
+//  vote required to exceeed the threshold is cast.  And the difference between propose and vote is distinguished by
+//  whether the --proposal=PROPOSAL_HASH flag is passed.  Later, we may want to split this into CONTRACT::COMMAND::propose,
```
RodrigoAD, line 6: We want to stick to the same process as EVM and Solana, where the multisig command tells us which step we are based on the proposal state

RodrigoAD:
Closing in favor of: https://github.com/smartcontractkit/chainlink-terra/pull/111

PR #117, opened 2022-02-15
bug/sc-29777 Add retries to gauntlet upload command

tateexon:


packages-ts/gauntlet-terra-contracts/src/commands/tooling/upload.ts
```diff
@@ -42,20 +43,29 @@ export default class UploadContractCode extends TerraCommand {
...
-      } catch (e) {
-        logger.error(`Error deploying ${contractName} code: ${e.message}`)
+          logger.success(`Contract ${contractName} code uploaded succesfully`)
+          contractReceipts[contractName] = res.tx
+          responses.push({
+            tx: res,
+            contract: null,
+          })
+        } catch (e) {
+          logger.error(`Error deploying ${contractName} Attempt ${retry + 1} Code: ${e.message}`)
```
RodrigoAD, line 62: Apparently there are some uploads that succeed. Could we give a more explanatory error message?
Also, could we ask confirmation from the user before every upload? (`await prompt(...)`)

tateexon: What other information would you like in the error message? It currently gives the contract, the number of times retried so far and the error message from the request. I really don't have any other information to put unless you want me to do some kind of swtich on the message from the request with different messaging based on whether it is a 400 or 404 or 502 or something like that.

I will get the prompt added.

tateexon: Changed the wording a bit and added the prompt.

packages-ts/gauntlet-terra-contracts/src/commands/tooling/upload.ts
```diff
@@ -42,20 +43,30 @@ export default class UploadContractCode extends TerraCommand {
...
+          logger.success(`Contract ${contractName} code uploaded succesfully`)
+          contractReceipts[contractName] = res.tx
+          responses.push({
+            tx: res,
+            contract: null,
+          })
+        } catch (e) {
+          logger.error(`Error deploying ${contractName} on attempt ${retry + 1} with the error: ${e.message}`)
+          if (maxRetry === retry + 1) {
+            throw e
```
RodrigoAD, line 65: Can we rethrow it like `throw new Error(e.response.data)`? Try to safely access the `data`

RodrigoAD (approved):

PR #119, opened 2022-02-15
Release v0.1.0 (contracts version)

krebernisak:


.github/workflows/codeql-analysis.yml
```diff
@@ -39,6 +39,11 @@ jobs:
     steps:
     - name: Checkout repository
       uses: actions/checkout@v2
+    - name: Set up (force) Go 1.17
+      uses: actions/setup-go@v1
+      with:
+        go-version: 1.17.2
```
jmank88, line 45: Minor, but they're up to 1.17.7 - maybe wildcard the patch?
```suggestion
        go-version: 1.17.x
```

krebernisak: Yeah, agreed. I'll add suggestions in a new PR on top of develop, shouldn't block this.

PR #128, opened 2022-02-17
pkg/terra: add Test_parseAttributes

jmank88:
https://app.shortcut.com/chainlinklabs/story/29944

### DIFF
--- a/pkg/terra/contract_reader.go
+++ b/pkg/terra/contract_reader.go
@@ -67,66 +67,153 @@ func (r *OCR2Reader) LatestConfig(ctx context.Context, changedInBlock uint64) (t
 
 	for _, event := range res.TxResponses[0].Logs[0].Events {
 		if event.Type == "wasm-set_config" {
-			return parseAttributes(event.Attributes)
+			cc, unknown, err := parseAttributes(event.Attributes)
+			if len(unknown) > 0 {
+				r.lggr.Warnf("wasm-set_config event contained unrecognized attributes: %v", unknown)
+			}
+			return cc, err
 		}
 	}
 	return types.ContractConfig{}, fmt.Errorf("No set_config event found for tx %s", res.TxResponses[0].TxHash)
 }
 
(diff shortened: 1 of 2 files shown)


pkg/terra/contract_reader.go
```diff
@@ -80,19 +82,19 @@ func parseAttributes(attrs []cosmosSDK.Attribute) (output types.ContractConfig,
 		case "latest_config_digest":
 			// parse byte array encoded as hex string
 			if err := HexToConfigDigest(value, &output.ConfigDigest); err != nil {
-				return types.ContractConfig{}, err
+				return types.ContractConfig{}, errors.Wrapf(err, "failed to parse attribute %q", key)
```
krebernisak, line 85: nit: we could have a helper:

```suggestion
				return attrError(key)
```

krebernisak (dismissed):

pkg/terra/contract_reader.go
jmank88: I actually had exactly that at first, but then the stack trace points to the helper. Can probably just add a proper error type to have it both ways.

archseer (dismissed):

pkg/terra/contract_reader.go
krebernisak: No worries, we can merge it as is :)

jmank88: np I've got some more changes incoming already

pkg/terra/contract_reader.go
```diff
@@ -67,66 +67,153 @@ func (r *OCR2Reader) LatestConfig(ctx context.Context, changedInBlock uint64) (t
...
+		return true
+	}
 	for _, attr := range attrs {
 		key, value := attr.Key, attr.Value
 		switch key {
 		case "latest_config_digest":
+			if !first(key) {
+				err = ErrAttrDupe(key)
+				return
+			}
```
krebernisak, line 102: Consider extracting this outside the `switch` statement, and have it declared just once under `for..range` instead of once per `case`.

krebernisak (approved):
Very nice!

PR #130, opened 2022-02-17
migrate terrad -> gauntlet

akhilchainani:


ops/terrad/terrad.go
```diff
@@ -27,92 +29,85 @@ const (
...
-
 type key struct {
 	Name    string
 	Address string
 }
 
-func New(ctx *pulumi.Context) (Terrad, error) {
-	// check if terrad is installed
-	_, err := exec.LookPath("terrad")
+type Deployer struct {
```
krebernisak, line 41: This should not be a migration/replacement but an addition as the first step.

For example, migrate `ops/terrad/terrad.go` as is to `ops/deployer/terrad.go` and add a separate `ops/deployer/guntlet.go`. Let's get both working first.

krebernisak (changes requested):

PR #132, opened 2022-02-18
Feature/29149 multisig group commands

sdrug:


packages-ts/gauntlet-terra-contracts/src/commands/contracts/multisig/updateMembers.ts
```diff
@@ -0,0 +1,66 @@
...
+}
+
+type ContractInput = {
+  add: Member[]
+  remove: string[]
+}
+
+const makeCommandInput = async (flags: any, args: any[]): Promise<CommandInput> => {
+  return {
+    add: flags.add?.split(','),
```
RodrigoAD, line 18: If we default these to an empty array `add: flags.add?.split(',') || []`, the CommandInput could be: 
```js
type CommandInput = {
  add: string[]
  remove: string[]
}
```
So we avoid questioning every time if they exist

packages-ts/gauntlet-terra-contracts/src/commands/contracts/multisig/updateMembers.ts
```diff
@@ -0,0 +1,66 @@
...
+const validateInput = (input: CommandInput): boolean => {
+  if (input.add && !input.add?.every((addr) => isValidAddress(addr))) {
+    throw new Error("One of provided 'add' addresses is not valid!")
+  }
+
+  if (input.remove && !input.remove?.every((addr) => isValidAddress(addr))) {
+    throw new Error("One of provided 'remove' addresses of not valid!")
+  }
+
+  if (!input.add && !input.remove) {
```
RodrigoAD, line 32: Here we could check that the length instead

packages-ts/gauntlet-terra-contracts/src/lib/multisig.ts
```diff
@@ -0,0 +1,4 @@
+export type Member = {
```
RodrigoAD, line 1: I don't expect this type list to grow much, let's bring this type over here: https://github.com/smartcontractkit/chainlink-terra/pull/132/files#diff-efaf39312302f9a5d3ac0b8232f61b9c01fe29526694c360400282cc5f779ee2R10

RodrigoAD (commented):
Looks really good, just a couple of comments

packages-ts/gauntlet-terra-contracts/src/commands/contracts/multisig/updateAdmin.ts
```diff
@@ -0,0 +1,45 @@
...
+  return true
+}
+
+const makeContractInput = async (input: CommandInput): Promise<ContractInput> => {
+  return {
+    admin: input.admin,
+  } as ContractInput
+}
+
+// yarn gauntlet cw4_group:update_admin --admin=<NEW_ADMIN_ADDRESS> <CONTRACT_ADDRESS>
```
krebernisak, line 33: @RodrigoAD where is the place to add these examples (in code)?

@sdrug why don't you also add some docs on these commands in `./docs`. It's empty currently but this is a good opportunity to start adding to it. Check out this rudimentary example from Solana: https://github.com/smartcontractkit/chainlink-solana/tree/develop/docs/multisig

```suggestion
```

packages-ts/gauntlet-terra-contracts/src/commands/contracts/multisig/updateAdmin.ts
```diff
@@ -0,0 +1,45 @@
+import { CATEGORIES } from '../../../lib/constants'
```
krebernisak, line 1: Can we please move this command to a `../cw4_group/...` subfolder, as this is the canonical name of the contract.

Also, why is this in the `gauntlet-terra-contracts` package and not in the `gauntlet-terra-cw-plus` package, which is the name (`cw-plus`) of the vendor source of contracts? cc @RodrigoAD

If there is a blocker that we can't do it now, let's at least be explicit in the folder structure and inline docs that this should be moved to another location.

packages-ts/gauntlet-terra-contracts/src/commands/contracts/multisig/updateMembers.ts
```diff
@@ -0,0 +1,66 @@
+import { CATEGORIES } from '../../../lib/constants'
```
krebernisak, line 1: Same comment as above on the location of this file.

packages-ts/gauntlet-terra-contracts/src/commands/contracts/multisig/updateMembers.ts
```diff
@@ -0,0 +1,66 @@
...
+  })
+
+  return {
+    add: membersToAdd || [],
+    remove: input.remove || [],
+  } as ContractInput
+}
+
+// yarn gauntlet cw4_group:update_members --add=<ADDRESS1_TO_ADD>,<ADDRESS2_TO_ADD> --remove=<ADDRESS3_TO_REMOVE> <CONTRACT_ADDRESS>
+// either --remove and --add can be omitted
```
krebernisak, line 54: Same comment on the docs as above.

packages-ts/gauntlet-terra-contracts/src/commands/contracts/multisig/updateAdmin.ts
RodrigoAD: This is blocked until we move abstract commands from `gauntlet-terra-contracts` to `gauntlet-terra`. Opening issue

RodrigoAD: https://github.com/smartcontractkit/chainlink-terra/issues/136

RodrigoAD (dismissed):
LGTM! Thanks @sdrug

RodrigoAD (approved):

PR #141, opened 2022-02-21
Add gauntlet e2e tests through accepting a proposal

tateexon:
Adds testing for more of the gauntlet commands
~Fixes the token:deploy to use a codeid in the codeid file provided if the hardcoded network one doesn't exist~
Cleans up an upload command issue where it can make all the upload calls too quickly which can cause false positive test failure.


tests/e2e/gauntlet_deployer.go
```diff
@@ -0,0 +1,370 @@
...
+
+// UpdateReportName updates the report name to be used by gauntlet on completion
+func UpdateReportName(reportName string, g *gauntlet.Gauntlet) {
+	g.NetworkConfig["REPORT_NAME"] = filepath.Join(utils.Reports, reportName)
+	err := g.WriteNetworkConfigMap(utils.Networks)
+	Expect(err).ShouldNot(HaveOccurred(), "Failed to write the updated .env file")
+}
+
+// GetInspectionResultsFromOutput parses the inpsectiond data from the output
+//  TODO we should really update the inspection command to just output json in the future
```
RodrigoAD, line 63: Can you open an issue/story with this? Maybe the json output could be the expected data vs. what we got onchain

tests/e2e/gauntlet_deployer.go
```diff
@@ -0,0 +1,370 @@
...
+
+// GetTxAddressFromReport gets the address from the typical place in the json report data
+func GetTxAddressFromReport(report map[string]interface{}) string {
+	return report["responses"].([]interface{})[0].(map[string]interface{})["tx"].(map[string]interface{})["address"].(string)
+}
+
+// DeployToken attempts to deploy the link token, currently does not work very often and I
+//  have never found a good reason why - Tate
+func (gd *GauntletDeployer) DeployToken() {
+	// TODO figure out why this never passes??? for a future pr
```
RodrigoAD, line 125: What's the exact issue here?

tests/e2e/gauntlet_deployer.go
```diff
@@ -0,0 +1,370 @@
...
+
+// ProposeConfig proposes the config
+func (gd *GauntletDeployer) ProposeConfig(ocr, proposalId, rddPath string) {
+	reportName := "propose_config"
+	UpdateReportName(reportName, gd.Cli)
+	_, err := gd.Cli.ExecCommandWithRetries([]string{
+		"ocr2:propose_config",
+		gd.Cli.Flag("version", gd.Version),
+		gd.Cli.Flag("rdd", rddPath),
+		gd.Cli.Flag("f", "1"),
```
RodrigoAD, line 291: `f` comes from RDD, so this shouldn't be necessary

tests/e2e/gauntlet_deployer.go
```diff
@@ -0,0 +1,370 @@
...
+func (gd *GauntletDeployer) ProposeOffchainConfig(ocr, proposalId, rddPath string) {
+	reportName := "propose_offchain_config"
+	gd.Cli.NetworkConfig["SECRET"] = gd.Cli.NetworkConfig["MNEMONIC"]
+	UpdateReportName(reportName, gd.Cli)
+	_, err := gd.Cli.ExecCommandWithRetries([]string{
+		"ocr2:propose_offchain_config",
+		gd.Cli.Flag("version", gd.Version),
+		gd.Cli.Flag("rdd", rddPath),
+		gd.Cli.Flag("proposalId", proposalId),
+		gd.Cli.Flag("offchainConfigVersion", "2"),
```
RodrigoAD, line 310: This is hardcoded on Gauntlet to be `2` always, so we can remove it

RodrigoAD (commented):
Looks really good, couple of comments

tests/e2e/gauntlet_deployer.go
tateexon: I started triaging into that again last night. I think I have figured it out and will push the fix with this pr. The deploy function is taking a codeIds argument but never using it in favor of just using hard coded ids for the cw20_base. 

I will add the ability to pull from a codeids file if the hardcoded network does not exist and add a prompt stating so when it does occur.

tests/e2e/gauntlet_deployer.go
tateexon: Created a story for this.

tests/e2e/gauntlet_deployer.go
tateexon: fixed

tests/e2e/gauntlet_deployer.go
tateexon: fixed

RodrigoAD (approved):

PR #145, opened 2022-02-22
Improvements after Multisig testing

RodrigoAD:
- Inspection gives more precise information
- Multisig afterExecute hook is executed
- Inspect Multisig command added
- Minor refactors

### DIFF
--- a/packages-ts/gauntlet-terra-contracts/src/commands/abstract/index.ts
+++ b/packages-ts/gauntlet-terra-contracts/src/commands/abstract/index.ts
@@ -171,7 +171,7 @@ export default class AbstractCommand extends TerraCommand {
   abstractQuery: AbstractExecute = async (params: any, address: string) => {
     logger.loading(`Calling ${this.opts.function} from contract ${this.opts.contract.id} at ${address}`)
     const result = await this.query(address, params)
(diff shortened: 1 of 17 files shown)


packages-ts/gauntlet-terra-contracts/src/commands/abstract/index.ts
```diff
@@ -171,7 +171,7 @@ export default class AbstractCommand extends TerraCommand {
   abstractQuery: AbstractExecute = async (params: any, address: string) => {
     logger.loading(`Calling ${this.opts.function} from contract ${this.opts.contract.id} at ${address}`)
     const result = await this.query(address, params)
-    logger.info(`Query finished with result: ${JSON.stringify(result)}`)
+    logger.debug(`Query finished with result: ${JSON.stringify(result)}`)
```
sdrug, line 174: I noticed that the result of the query call is not shown in the console unless you enable debug flag. Maybe still let it be info?

RodrigoAD: The inspection command uses this underneath, if we let it into `info` the command gets too verbose. Debug is fine, can be enabled with `DEBUG=true` flag

sdrug (approved):

PR #148, opened 2022-02-23
proxy cmds uses args[0] and pull aggregator from RDD

momentmaker:


packages-ts/gauntlet-terra-contracts/src/commands/contracts/proxy_ocr2/deploy.ts
```diff
@@ -13,8 +14,12 @@ type ContractInput = {
 }
 
 const makeCommandInput = async (flags: any, args: string[]): Promise<CommandInput> => {
+  const rdd = getRDD(flags.rdd)
+  const contract = args[0]
```
RodrigoAD, line 18: Thanks @momentmaker ! Could you add it to propose and accept proxy commands as well?

momentmaker: @RodrigoAD sure done :)

packages-ts/gauntlet-terra-contracts/src/commands/contracts/proxy_ocr2/confirmContract.ts
```diff
@@ -13,8 +14,12 @@ type ContractInput = {
 }
 
 const makeCommandInput = async (flags: any, args: string[]): Promise<CommandInput> => {
+  const rdd = getRDD(flags.rdd)
+  const contract = args[0]
+  const aggregator = rdd.contracts[contract]
```
RodrigoAD, line 19: We don't need RDD on these commands, just the contract address.

momentmaker: But we still want to verify that the contract address exists in the RDD right?

momentmaker: nvm there was a bug of passing in the whole template into `address` which gives validation error

momentmaker:
Ahh it's not set in `testnet-bombay.json`

```
{
  "flags": 39112,
  "ocr2": 45851,
  "access_controller": 39116,
  "deviation_flagging_validator": 39115,
  "cw3_flex_multisig": 36059,
  "cw4_group": 36895,
  "cw20_base": 41738
}
```

krebernisak (approved):

PR #151, opened 2022-02-24
Move abstract command features into gauntlet-terra package

reductionista:
This moves the abstract commands from @chainlink/gauntlet-terra-contracts into @chainlink/gauntlet-terra package.

The two criteria I followed while refactoring were:
1.  AbstractCommand, AbstractInstruction, InspectInstruction, makeAbstractCommand, instructionToCommand, and instructionToInspectCommand should all be moved to @chainlink/terra-gauntlet
2. There shouldn't be any contract-specific code in @chainlink/terra-gauntlet, anything referencing specific contracts (ocr, access_controller, etc. should stay in @chainlink/terra-gauntlet-contracts

These were difficult to satisfy simultaneously without introducing circular dependencies between the two packages, since the commands in (1.) all call getContract(), and reference CONTRACT_LIST as well as schema abi stuff.
I moved most of the schema code into @chainlink/terra-gauntlet, since that was more general anyway.  But getContract() and CONTRACT_LIST were trickier.  The solution I implemented for eliminating these circular dependencies was to have these two parameters be passed from terra-guantlet-contracts upon initialization into an AbstractTools class in terra-gauntlet, which contains all of the necessary functions in (1).  ocr deploy, addAccess, and removeAccess are the three I think are working now.  If that looks good, I'll go ahead and copy the changes in imports and call signatures to all the other custom commands.

Also: ajv library requires "use strictNullChecks" to be set in package.json.  We had this set for the package it was in, but not for gauntlet-terra.  Adding this compiler flag introduced a lot of type errors, but they should all be fixed now.


packages-ts/gauntlet-terra-contracts/src/commands/contracts/access_controller/addAccess.ts
```diff
@@ -1,6 +1,7 @@
 import { AccAddress } from '@terra-money/terra.js'
-import { AbstractInstruction, instructionToCommand } from '../../abstract/executionWrapper'
+import { abstract, AbstractInstruction } from '../..'
```
RodrigoAD, line 2: We should import this from `@chainlink/gauntlet-terra`

packages-ts/gauntlet-terra-contracts/src/commands/contracts/cw4_group/deploy.ts
```diff
@@ -1,6 +1,7 @@
 import { CATEGORIES } from '../../../lib/constants'
 import { isValidAddress } from '../../../lib/utils'
-import { AbstractInstruction, instructionToCommand } from '../../abstract/executionWrapper'
+import { CONTRACT_LIST, getContract } from '../../../lib/contracts'
```
RodrigoAD, line 3: This is not used, right?

packages-ts/gauntlet-terra-contracts/src/commands/contracts/access_controller/addAccess.ts
reductionista: We can't instantiate `AbstractTools` in `@gauntlet-terra`, since it depends on `CONTRACT_LIST`, which is specific to `@gauntlet-terra-contracts`.  We either have to instantiate a different copy of it in every file here, or instantiate it once and share it within the package which is the option I went with here.

Moving `AbstractInstruction` into `@gauntlet-terra` is possibility, but would be messy since it also depends on
 `CONTRACT_LIST`.  In the present PR, a templated version of it `AbstractInstructionTemplate` is defined in `@gauntlet-terra`, and then the more specific type `AbstractInstruction` is built from that in `@guantlet-terra-contracts` and shared among the files.  We can change that, but it would mean importing `CONTRACT_LIST` in every file, just so we can fill in that template parameter.

One thing that might make this a bit cleaner is if we added a small `abstract.ts` to `@gauntlet-terra-contract`, for managing the abstract imports from `@gauntlet-terra`.  Presently, I included this in `@gauntlet-terra-contract/src/commands/index.ts`, but it probably deserves its own file.

packages-ts/gauntlet-terra-contracts/src/commands/contracts/cw4_group/deploy.ts
reductionista: Correct.  Originally, I was doing something similar to what you suggested in the other comment... which required importing `CONTRACT_LIST` everywhere.  After splitting it into `AbstractInstructionTemplate` and `AbstractInstruction`, that made it unnecessary to import `CONTRACT_LIST` in most of the files... must have missed this one while going through and removing all of the imports.  If we do decide to import `AbstractInstruction` directly from `@gauntlet-terra`, then we'll have to keep this import and add a lot more of them.

PR #155, opened 2022-02-25
change env naming convention

momentmaker:


tateexon (commented):
I think you will need to update the constants here https://github.com/smartcontractkit/chainlink-terra/blob/ddd0db809d7e89bedd7edf8bb196b72ed28dae77/packages-ts/gauntlet-terra-contracts/src/lib/constants.ts or you will break some things like the gauntlet `token:deploy` command.

momentmaker:
@tateexon awesome! thank you for the heads up. I wasn't aware of those. could you please review again?

packages-ts/gauntlet-terra-contracts/codeIds/bombay-testnet.json
```diff
@@ -1,9 +0,0 @@
-{
```
tateexon, line 1: This just needed renamed, not deleted right?

tateexon (changes requested):

packages-ts/gauntlet-terra-contracts/codeIds/bombay-testnet.json
momentmaker: oops. I had renamed it but git add didn't automatically pick it up

tateexon (approved):
Looks good as far as I can tell. I at least don't know of anything else it could break.

PR #157, opened 2022-02-28
Token commands + Improvements

RodrigoAD:
- OCR Inspection command checks how much it owes to the transmitters and compares with its current balance
- Transfer Token command now uses abstract command
- Send LUNA command added.
- `beforeExecute` hook added

PR #159, opened 2022-03-01
30252 tx simulation

sdrug:


packages-ts/gauntlet-terra-contracts/src/commands/abstract/executionWrapper.ts
```diff
@@ -14,12 +14,15 @@ export interface AbstractInstruction<Input, ContractInput> {
   validateInput: (input: Input) => boolean
   makeContractInput: (input: Input) => Promise<ContractInput>
   afterExecute?: (response: Result<TransactionResponse>) => any
+  isSimulateBeforeExecution?: boolean
```
RodrigoAD, line 17: I think we will always want the simulation, can't think of a case where we don't want it

packages-ts/gauntlet-terra-contracts/src/commands/abstract/index.ts
```diff
@@ -192,6 +192,20 @@ export default class AbstractCommand extends TerraCommand {
     }
   }
 
+  simulateExecute = async () => {
```
RodrigoAD, line 195: Let's return the estimated gas here

packages-ts/gauntlet-terra-contracts/src/commands/contracts/ocr2/proposal/acceptProposal.ts
```diff
@@ -63,6 +63,7 @@ const instruction: AbstractInstruction<CommandInput, ContractInput> = {
   validateInput: validateInput,
   makeContractInput: makeContractInput,
   afterExecute,
+  isSimulateBeforeExecution: true,
```
RodrigoAD, line 66: If we leave the simulation active by default, this shouldn't be necessary

packages-ts/gauntlet-terra/src/commands/internal/terra.ts
```diff
@@ -134,4 +138,21 @@ export default abstract class TerraCommand extends WriteCommand<TransactionRespo
 
     return this.wrapResponse(res)
   }
+
+  async simulate(address, input): Promise<Number> {
+    const msg = new MsgExecuteContract(this.wallet.key.accAddress, address, input)
```
RodrigoAD, line 143: Simulate should accept a `signer` param. The simulation should happen using that signer. That will be useful for the multisig case. 
```suggestion
 async simulate(signer: AccAddress, address: AccAddress, input: any): Promise<Number> {
    const msg = new MsgExecuteContract(signer, address, input)
```

packages-ts/gauntlet-terra/src/commands/internal/terra.ts
```diff
@@ -134,4 +138,21 @@ export default abstract class TerraCommand extends WriteCommand<TransactionRespo
...
+
+    const signerData: SignerData = {
+      sequenceNumber: await this.wallet.sequence(),
+      publicKey: this.wallet.key.publicKey,
```
RodrigoAD, line 147: We'll need a way to get the `signer` sequence number
The public key should be the signer

packages-ts/gauntlet-terra-contracts/src/commands/abstract/index.ts
```diff
@@ -192,6 +192,20 @@ export default class AbstractCommand extends TerraCommand {
...
+    if (this.opts.action !== TERRA_OPERATIONS.EXECUTE) {
+      logger.info('Skipping tx simulation for non-execute operation')
+      return
+    }
+
+    const address = this.args[0]
+    const params = this.params
+    logger.loading(`Executing tx simulation for ${this.opts.contract.id}:${this.opts.function} at ${address}`)
+
+    const estimatedGas = await this.simulate(address, params)
```
RodrigoAD, line 205: As abstract command will always use the default loaded wallet, we can pass the `this.wallet.key.accAddress` signer param here

packages-ts/gauntlet-terra/src/commands/internal/terra.ts
```diff
@@ -22,8 +23,11 @@ export default abstract class TerraCommand extends WriteCommand<TransactionRespo
   provider: LCDClient
   contracts: string[]
   public codeIds: CodeIds
+
   abstract execute: () => Promise<Result<TransactionResponse>>
   abstract makeRawTransaction: (signer: AccAddress) => Promise<MsgExecuteContract>
+
+  simulateExecute: () => any
```
RodrigoAD, line 30: Is this necessary?

RodrigoAD: With [this change](https://github.com/smartcontractkit/chainlink-terra/pull/157/files#diff-48dc1bdb44afa06fa4ff99d7b4fef1f6dc28a090ead3796e0c5365bd711d0c29R24) shouldn't be necessary

packages-ts/gauntlet-terra/src/commands/internal/terra.ts
```diff
@@ -139,13 +144,15 @@ export default abstract class TerraCommand extends WriteCommand<TransactionRespo
     return this.wrapResponse(res)
   }
 
-  async simulate(address, input): Promise<Number> {
-    const msg = new MsgExecuteContract(this.wallet.key.accAddress, address, input)
+  async simulate(signer: AccAddress, contractAddress: AccAddress, input: any): Promise<Number> {
```
RodrigoAD, line 147: I think would make sense this to accept a list of messages instead of a contract and an input, something like the [signAndSend](https://github.com/smartcontractkit/chainlink-terra/blob/develop/packages-ts/gauntlet-terra/src/commands/internal/terra.ts#L68) function.
This would the function also easier to use from the multisig (or any other consumer)
```suggestion
  async simulate(signer: AccAddress, msgs: MsgExecuteContract[]): Promise<Number> {
  ...
  const tx = await this.wallet.createTx({
      msgs,
  })
```

packages-ts/gauntlet-terra-cw-plus/src/commands/multisig.ts
```diff
@@ -31,6 +31,8 @@ export const wrapCommand = (command) => {
 
     makeRawTransaction = async (signer: AccAddress, state?: State) => {
       const message = await this.command.makeRawTransaction(this.multisig)
+      const estimatedGas = await this.command.simulate(this.multisig, [message])
+      logger.info(`Tx simulation successful: estimated gas usage is ${estimatedGas}`)
```
RodrigoAD, line 35: Nice, it was exactly this. 
The message is not entirely accurate though, this tx won't be executed directly, it will be wrapped into the multisig proposal, so the gas usage should be from the multisig tx simulation

RodrigoAD (dismissed):
Looks good to me!

krebernisak (dismissed):

RodrigoAD:
Merge if you think is ready @sdrug

RodrigoAD (approved):

PR #160, opened 2022-03-01
Fix/initialize flow proposal id and auto write

sdrug:


packages-ts/gauntlet-terra-contracts/src/commands/contracts/ocr2/initialize.flow.ts
```diff
@@ -34,7 +36,7 @@ export default class OCR2InitializeFlow extends FlowCommand<TransactionResponse>
       },
       {
         name: 'Change RDD',
-        exec: this.showRddInstructions,
+        exec: () => this.overwriteRDD(flags),
```
RodrigoAD, line 39: Why this change? Gauntlet currently only reads from RDD, I don't know if this approach is the correct one to update RDD

sdrug: @RyanRHall what was your reasoning behind adding it?
I also agree with Rodrigo that the gauntlet should refrain from implicitly overwriting any configuration files.

PR #162, opened 2022-03-01
Add Ocr2 Proxy E2E Test

tateexon:
Also bumped up the concurrency from 2 to 3 so we hopefully don't increase run time in ci by much if any.

### DIFF
--- a/tests/e2e/smoke/ocr2_proxy_test.go
+++ b/tests/e2e/smoke/ocr2_proxy_test.go
@@ -1,0 +1,57 @@
+package smoke_test
+
+import (
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"github.com/smartcontractkit/chainlink-terra/tests/e2e"
+	"github.com/smartcontractkit/chainlink-terra/tests/e2e/common"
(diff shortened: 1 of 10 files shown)


tests/e2e/smoke/ocr2_proxy_test.go
```diff
@@ -0,0 +1,57 @@
...
+	tc "github.com/smartcontractkit/chainlink-terra/tests/e2e/smoke/common"
+	"github.com/smartcontractkit/integrations-framework/actions"
+)
+
+var _ = Describe("Terra OCRv2 Proxy @ocr_proxy", func() {
+	var state *tc.OCRv2State
+
+	BeforeEach(func() {
+		state = &tc.OCRv2State{}
+		By("Deoloying the cluster", func() {
```
aalu1418, line 19:
```suggestion
		By("Deploying the cluster", func() {
```

aalu1418 (approved):
LGTM! (just a spelling change that i noticed)

tests/e2e/smoke/ocr2_proxy_test.go
tateexon: Good catch! I copy pasted from another test and it had the issue so I fixed it in all the copy pasted areas and found some others.

aalu1418 (approved):
🎉

PR #165, opened 2022-03-02
Support negative observations

connorwstein:


pkg/terra/marshal_signed_int.go
```diff
@@ -0,0 +1,53 @@
...
+		val.Sub(maxUint, val)
+		val.Neg(val)
+	}
+	return val, nil
+}
+
+// ToBytes converts *big.Int o into bytes as a big-endian signed
+// integer of size numBytes
+func ToBytes(o *big.Int, numBytes uint) ([]byte, error) {
+	negative := o.Sign() < 0
```
aalu1418, line 33: should this have validation that the big.Int fits in the expected number of bytes?

like in libocr: https://github.com/smartcontractkit/libocr/blob/449836e6cfec324fbd6781e9a4d4b953840cdf0b/offchainreporting2/reportingplugin/median/observation.go#L21

pkg/terra/marshal_signed_int.go
```diff
@@ -0,0 +1,53 @@
...
+// ToInt interprets bytes s as a big-endian signed integer
+// of size numBytes.
+func ToInt(s []byte, numBytes uint) (*big.Int, error) {
+	if uint(len(s)) != numBytes {
+		return nil, fmt.Errorf("invalid int length: expected %d got %d", numBytes, len(s))
+	}
+	val := (&big.Int{}).SetBytes(s)
+	numBits := numBytes * 8
+	// 2**(numBits-1) - 1
+	maxPositive := big.NewInt(0).Sub(big.NewInt(0).Lsh(big.NewInt(1), numBits-1), big.NewInt(1))
```
aalu1418, line 18: could separate this into another helper func so we can use it in both functions

aalu1418 (changes requested):
Also just a general question: where should we store this helper?

if we wanted to use this in chainlink-solana...would the imports get weird?

jmank88 (dismissed):

connorwstein:
> Also just a general question: where should we store this helper?
> 
> if we wanted to use this in chainlink-solana...would the imports get weird?

I'll try to just PR this to OCR

aalu1418 (approved):
🎉

PR #166, opened 2022-03-02
Gauntlet sec improvements

RodrigoAD:
- Added user confirmation with RDD checks in ownership commands
- Offchain config can be inspected without digest. Will get the offchain config from the tx log


packages-ts/gauntlet-terra-contracts/src/lib/provider.ts
```diff
@@ -0,0 +1,28 @@
...
+        {
+          key: 'tx.height',
+          value: String(block),
+        },
+      ],
+      'pagination.offset': String(offset),
+    })
+    return txs.txs.concat(await getBlockTxs(search, block, offset + 100))
+  } catch (e) {
+    logger.debug('No more txs in block')
```
krebernisak, line 24: We should probably swallow only a specific error (no more results) while rethrowing other potential errors.

packages-ts/gauntlet-terra-contracts/src/commands/contracts/ocr2/inspection/inspect.ts
```diff
@@ -45,12 +50,13 @@ const makeInput = async (flags: any, args: string[]): Promise<ContractExpectedIn
...
       transmissionPaymentGjuels: info.billing.transmissionPaymentGjuels,
     },
+    offchainConfig: getOffchainConfigInput(rdd, contract),
   }
 }
 
 const makeInspectionData = () => async (input: ContractExpectedInfo): Promise<ContractExpectedInfo> => input
 
-const makeOnchainData = (query: Query) => async (
+const makeOnchainData = (query: Query, search: Search) => async (
```
krebernisak, line 59: Anyone reading this first question will be what's the difference between a `Query` and a `Search` 🤔

krebernisak: Why don't we have a `provider` as an input param and then do:

- `provider.wasm.contractQuery({...})`
- `provider.tx.search({...})`

It would make the code a lot more understandable IMO.

packages-ts/gauntlet-terra-contracts/src/commands/abstract/executionWrapper.ts
```diff
@@ -3,14 +3,21 @@ import { Result } from '@chainlink/gauntlet-core'
...
 import { AccAddress } from '@terra-money/terra.js'
 import { logger, prompt } from '@chainlink/gauntlet-core/dist/utils'
+import { Query } from './inspectionWrapper'
 
 export type BeforeExecutionContext<Input, ContractInput> = {
   input: Input
   contractInput: ContractInput
   id: string
   contract: string
+  query: Query
```
krebernisak, line 13: This `query` seems off and would rather see a full `provider` here.

packages-ts/gauntlet-terra-contracts/src/commands/abstract/executionWrapper.ts
```diff
@@ -3,14 +3,21 @@ import { Result } from '@chainlink/gauntlet-core'
 import { TerraCommand, TransactionResponse } from '@chainlink/gauntlet-terra'
 import { AccAddress } from '@terra-money/terra.js'
 import { logger, prompt } from '@chainlink/gauntlet-core/dist/utils'
+import { Query } from './inspectionWrapper'
 
 export type BeforeExecutionContext<Input, ContractInput> = {
```
krebernisak, line 8: Is there any other type of context? Why don't we just call it `ExecutionContext`?

packages-ts/gauntlet-terra-contracts/src/commands/abstract/executionWrapper.ts
```diff
@@ -21,7 +28,7 @@ export interface AbstractInstruction<Input, ContractInput> {
   makeInput: (flags: any, args: string[]) => Promise<Input>
   validateInput: (input: Input) => boolean
   makeContractInput: (input: Input) => Promise<ContractInput>
-  beforeExecute?: (context: BeforeExecutionContext<Input, ContractInput>) => () => Promise<void>
+  beforeExecute?: BeforeExecute<Input, ContractInput>
   afterExecute?: (response: Result<TransactionResponse>) => any
```
krebernisak, line 32: Feel free to also type this function, something like:

```
export type AfterExecute<Input, ContractInput> = (
  context: ExecutionContext<Input, ContractInput>,
  response: Result<TransactionResponse>,
) => Promise<any>
```

Additionally, not too fond of these `<Input, ContractInput>` types overall but no suggestions yet.

packages-ts/gauntlet-terra-contracts/src/commands/abstract/executionWrapper.ts
```diff
@@ -3,14 +3,21 @@ import { Result } from '@chainlink/gauntlet-core'
...
 export type BeforeExecutionContext<Input, ContractInput> = {
   input: Input
   contractInput: ContractInput
   id: string
   contract: string
+  query: Query
+  flags: any
 }
 
+export type BeforeExecute<Input, ContractInput> = (
```
krebernisak, line 17: Why is this a function, returning a function, returning a promise?

Makes more sense for the `BeforeExecute` type to be a simple function returning a promise?

packages-ts/gauntlet-terra-contracts/src/commands/contracts/ownership/acceptOwnership.ts
```diff
@@ -1,21 +1,32 @@
...
+    logger.warn('No RDD flag provided. Accepting ownership without RDD check')
+    logger.info(`Accepting Ownership Transfer of contract with current owner ${currentOwner} to new owner `)
+    await prompt('Continue?')
+    return
+  }
+  const contract = getContractFromRDD(getRDD(context.flags.rdd), context.contract)
+  logger.info(`Accepting Ownership Transfer of contract of type "${contract.type}":
+    - Contract: ${contract.address} ${contract.description ? '- ' + contract.description : ''}
+    - Current Owner: ${currentOwner}
+    - Next Owner (Current signer): ${signer}
```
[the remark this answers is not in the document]
RodrigoAD: Also the contract type (proxy, contract, ac), and basically verifies that the contract actually exists on RDD

packages-ts/gauntlet-terra/src/commands/middlewares.ts
```diff
@@ -37,7 +37,7 @@ export const withWallet: Middleware = async (c: TerraCommand, next: Next) => {
   }
 
   c.wallet = c.provider.wallet(key)
-  logger.info(`Operator address is ${c.wallet.key.accAddress}`)
+  logger.debug(`Operator address is ${c.wallet.key.accAddress}`)
```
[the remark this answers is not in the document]
RodrigoAD: Multisig is currently wrapping 3 commands, and this can become noisy, as will be prompted 3 times in that case. I guess we could move the log somewhere else

packages-ts/gauntlet-terra-contracts/src/lib/rdd.ts
```diff
@@ -21,3 +21,33 @@ export const getRDD = (path: string, fileDescription: string = 'RDD'): any => {
...
+      return {
+        type,
+        contract: content,
+        address,
+        ...((type === CONTRACT_TYPES.CONTRACT || type === CONTRACT_TYPES.PROXY) && { description: content.name }),
+      }
+    }
+    return agg
+  }, {} as RDDContract)
+}
```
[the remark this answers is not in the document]
RodrigoAD: Not sure about moving it to `gauntlet-terra`, is quite contract specific no? 

I imagine this function to change in the future, it can become quite useful. We can update it when we have other use cases for it

packages-ts/gauntlet-terra-contracts/src/lib/provider.ts
RodrigoAD: Makes sense

packages-ts/gauntlet-terra-contracts/src/commands/abstract/executionWrapper.ts
RodrigoAD: The instruction command will provide the context in what the execution will happen. Then is responsibility of the command executor to actually use the hook.
IE, multisig command don't have the context of what's happening in the underlying commands, just wants to trigger the hook

packages-ts/gauntlet-terra-contracts/src/lib/rdd.ts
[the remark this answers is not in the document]
krebernisak: That's why it needs more thought, and feel free to leave it here s is. 

But IMO high-level ideas like `RDDContract` & `getContractFromRDD` (not talking about this specific impl) is a core thing and needs to be moved to the core level.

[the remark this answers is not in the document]
RodrigoAD: Moved to `gauntlet-terra`, and agree, this should go into a common place. Opening an issue on it

packages-ts/gauntlet-terra-contracts/src/lib/provider.ts
RodrigoAD: Logs the error if it's not what we expect.

packages-ts/gauntlet-terra-contracts/src/commands/contracts/ocr2/inspection/inspect.ts
RodrigoAD: With provider works like a charm

krebernisak (approved):

PR #169, opened 2022-03-03
unit test: reverted payments do not change owed balance

aalu1418:


krebernisak (approved):
Nice, you're getting the hang of it :)

PR #172, opened 2022-03-04
Fee coin bounds check

connorwstein:


crates/owned/src/lib.rs
```diff
@@ -101,7 +101,7 @@ impl<'a> Auth<'a> {
         self.0.save(deps.storage, &state)?;
 
         Ok(Response::default().add_event(
-            Event::new("ownership_transfer_requested")
+            Event::new("ownership_transfer_accepted")
```
connorwstein, line 104: the other event name actually made sense to me, only this one seems off (looks like a copy paste error)

jmank88 (approved):

jmank88 (approved):

PR #173, opened 2022-03-04
include chain_id length in config digest

aalu1418:


contracts/ocr2/src/state.rs
```diff
@@ -158,7 +158,8 @@ pub fn config_digest_from_data(
 ) -> [u8; 32] {
     use blake2::{Blake2s, Digest};
     let mut hasher = Blake2s::default();
-    hasher.update(chain_id.as_bytes());
+    hasher.update((chain_id.len() as u8).to_be_bytes());
```
connorwstein, line 161: lgtm - I think we just want that overflow check on and off chain

connorwstein (approved):

PR #174, opened 2022-03-04
Add instructions for running the e2e tests to the readme

tateexon:


README.md
```diff
@@ -19,3 +19,23 @@ This repository is a monorepo of the various components required for Chainlink o
 
     # Then run
     asdf install
+
+# Running e2e tests
```
sdrug, line 23: Could you please add the links you shared in the slack channel like where to go and what to do in order to configure AWS access

krebernisak, line 23: Going forward I'd prefer all long-form docs to go under `./docs` subfolder, markdown files in the appropriate section.

Do you mind moving those to `./docs` and actually linking to them from a README. Ideally a README from the `tests/e2e` subfolder.

krebernisak (commented):
Awesome @tateexon!

Do you mind documenting the same for Solana 🙏

README.md
tateexon: Updated to this formatting

krebernisak (dismissed):

krebernisak (approved):

Makefile
```diff
@@ -63,3 +63,6 @@ test_migration:
 
 test_gauntlet:
 	SELECTED_NETWORKS=localterra NETWORK_SETTINGS=$(shell pwd)/tests/e2e/networks.yaml ginkgo --focus=@gauntlet tests/e2e/smoke
+
```
krebernisak, line 66: There is a `ginkgo` install command in the Makefile.

Should we consider rather adding `ginkgo` to our NIX & asdf envs and incentivize usage of those envs instead of adding manual installation steps?

Notice: also applies to similar Solana PR

tateexon: We could add ginkgo to asdf but there is only a plugin for the `go install` and not the three `go get` libraries we also need, so we would be separating two things that should be ran together. 

I suppose we could put together an asdf plugin that does the `go get` commands along with the `go install`. I will take a look into it to see how hard it would be because we use this same ginkgo setup in multiple repos around the company so it might be good to just push it out to a plugin.

krebernisak: Yeah, try to figure out if we can make some improvements by adding it to our evns, but the current PR/setup is perfectly fine if this turns out to be complex.

PR #175, opened 2022-03-04
packages-ts/terra-sdk

### LINKED ISSUE #163: Explicitly declare minimum required Node.js version for any project that uses Node.js.
Neither `gauntlet-terra` nor `gauntlet-terra-contracts` give a minimum required Node.js version. The recommendation is that a minimum required Node.js version be given for any JavaScript or TypeScript project that uses Node.js.

jmank88:
https://app.shortcut.com/chainlinklabs/story/31300/client-to-stream-on-chain-events-for-data-products-analysis

Closes #163


archseer:
Quite nice 👍 I'd rename the directory to terra-sdk? We should use a similar name on Solana too (so the packages end up being chainlink/solana-sdk, terra-sdk)

archseer:
I'd recommend splitting out gauntlet to /gauntlet too, that way the workspace can be built without private dependencies.

jmank88:
@archseer  Is this what you mean?
```
/gauntlet/gauntlet-terra-*
/packages-ts/terra-sdk
```
I think I broke it...

gauntlet/gauntlet-terra-cw-plus/package.json
```diff
@@ -14,16 +14,16 @@
...
     "test:coverage": "yarn test --collectCoverage",
     "test:ci": "yarn test --ci",
-    "lint:format": "yarn prettier --check ./src",
-    "format": "yarn prettier --write ./src",
+    "lint:format": "yarn prettier --check src",
+    "format": "yarn prettier --write src",
     "clean": "rm -rf ./dist/ ./bin/",
     "build": "yarn clean && tsc",
-    "bundle": "yarn build && pkg ."
+    "bundle": "yarn build && pkg "
```
krebernisak, line 26: Not sure if the trailing dot is required, but now there is a trailing space here:

```suggestion
    "bundle": "yarn build && pkg"
```

gauntlet/gauntlet-terra/package.json
```diff
@@ -14,16 +14,16 @@
...
     "test:coverage": "yarn test --collectCoverage",
     "test:ci": "yarn test --ci",
-    "lint:format": "yarn prettier --check ./src",
-    "format": "yarn prettier --write ./src",
+    "lint:format": "yarn prettier --check src",
+    "format": "yarn prettier --write src",
     "clean": "rm -rf ./dist/ ./bin/",
     "build": "yarn clean && tsc",
-    "bundle": "yarn build && pkg ."
+    "bundle": "yarn build && pkg "
```
krebernisak, line 26:
```suggestion
    "bundle": "yarn build && pkg"
```
```suggestion
    "bundle": "yarn build && pkg "
```

krebernisak, line 26:
```suggestion
    "bundle": "yarn build && pkg"
```
```suggestion
    "bundle": "yarn build && pkg "
```

tsconfig.json
```diff
@@ -4,13 +4,13 @@
   "include": [],
   "references": [
```
krebernisak, line 5: Should also reference the `terra-sdk` package?

package.json
```diff
@@ -10,11 +10,11 @@
     "typescript",
     "cli"
   ],
-  "main": "packages-ts/gauntlet-terra-contracts/dist/index.js",
+  "main": "gauntlet/gauntlet-terra-contracts/dist/index.js",
   "bin": "packages-ts/gauntlet-terra-contracts/dist/index.js",
```
krebernisak, line 14: Missed a line here!

gauntlet/gauntlet-terra-contracts/package.json
```diff
@@ -14,17 +14,17 @@
...
     "test:ci": "yarn test --ci",
-    "lint:format": "yarn prettier --check ./src",
-    "format": "yarn prettier --write ./src",
+    "lint:format": "yarn prettier --check src",
+    "format": "yarn prettier --write src",
     "formatStaged": "node ./scripts/formatStaged.js",
     "clean": "rm -rf ./dist/ ./bin/",
     "build": "yarn clean && tsc",
-    "bundle": "yarn build && pkg ."
+    "bundle": "yarn build && pkg "
```
krebernisak, line 27:
```suggestion
    "bundle": "yarn build && pkg"
```

krebernisak, line 27:
```suggestion
    "bundle": "yarn build && pkg"
```

packages-ts/terra-sdk/src/index.ts
```diff
@@ -0,0 +1,57 @@
+import { WebSocketClient } from '@terra-money/terra.js';
+import { TxLog, Int } from '@terra-money/terra.js';
+
+export class Chainlink {
```
krebernisak, line 4: This class could be renamed as is specific to the OCR2 protocol.

packages-ts/terra-sdk/src/index.ts
```diff
@@ -0,0 +1,57 @@
...
+    private _wsClient: WebSocketClient;
+
+    constructor(
+        readonly url: string, //TODO or pass own client?
```
krebernisak, line 8: Yeah, think that passing in a client is a better & more flexible approach.

jest.config.js
```diff
@@ -1,6 +1,17 @@
...
   rootDir: '.',
   projects: [
+    {
+      displayName: 'terra-sdk',
+      preset: 'ts-jest',
+      testEnvironment: 'node',
+      testMatch: ['<rootDir>/packages-ts/terra-sdk/**/*.test.ts'],
+      globals: {
+        'ts-jest': {
+          tsconfig: '<rootDir>/packages-ts/terra-sdk/tsconfig.json',
```
krebernisak, line 11: IMO we should have one main TS (yarn) workspace for this monorepo and all! packages (something published) should be listed flat under `./packges-ts`.

This will lower the complexity of using, finding, and actually implementing packages. They are all in the same place, properly configured once.

There can be other TS projects in the monorepo outside `packages-ts` but those are not published packages (dependencies of other projects), but examples, tests & projects that are used directly.

Here is an example complex monorepo that has a similar structure:

- All TS published packages under ./packages: https://github.com/ethereum-optimism/optimism/tree/develop/packages
- For non-published packages, there could be other subfolders with TS code, for example:https://github.com/ethereum-optimism/optimism/tree/develop/integration-tests

krebernisak (dismissed):

jest.config.js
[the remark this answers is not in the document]
jmank88: I un-did the previous refactor so things are back under `packages-ts/` again.

packages-ts/terra-sdk/src/index.ts
```diff
@@ -0,0 +1,55 @@
...
+        this._wsClient = client;
+    }
+
+    public start() {
+        this._wsClient.start()
+    }
+
+    public destroy() {
+        this._wsClient.destroy()
+    }
```
boxhock, line 17: If `_wsClient` is passed in as a param when making this class, shouldn't we expect that the caller takes care of start/destroy?

boxhock (changes requested):
Had a couple of comments. Would also recommend implementing prettier or some linting to give a more consistent code styling.

packages-ts/terra-sdk/src/index.ts
jmank88: Ah yes - just switched the constructor from url, so this can be simpler now.

packages-ts/terra-sdk/package.json
```diff
@@ -0,0 +1,48 @@
...
+    "README.md"
+  ],
+  "sideEffects": false,
+  "main": "dist/cjs/index.js",
+  "module": "dist/esm/index.js",
+  "types": "dist/esm/index.d.ts",
+  "exports": {
+    "import": "./dist/esm/index.js",
+    "require": "./dist/cjs/index.js"
+  },
```
krebernisak, line 25: @HenryNguyen5 hey hey, we could use your expertise here! 
We're trying to simply package this simple package :)

Could you help us by reviewing and suggesting the proper configs to do so? We'd love to support browsers and multiple module options, but for now, we're just looking to support the usual general case and get this over quickly.

Any suggestions?

packages-ts/terra-sdk/src/index.ts
krebernisak: Yeah, the suggestion makes sense and the API needs more thought as we didn't have the time currently.

That being said, it should be ok as these are public methods currently (`start/destroy`) and not called internally. So it's up to the user to use them if desired.

packages-ts/terra-sdk/package.json
archseer: Solana could use some help on this too: https://github.com/smartcontractkit/chainlink-solana/pull/224

These might be of help, I compared tsconfigs for 
- [Solana](https://github.com/solana-labs/solana-program-library/tree/master/token/js)
- [Anchor](https://github.com/project-serum/anchor/tree/master/ts)
- [SaberSwap](https://github.com/saber-hq/saber-common/tree/master/packages/stableswap-sdk)

The last one seemed the cleanest but it imported a separate base tsconfig from somewhere else.

archseer: We could offer only ESM and offer CJS later if necessary https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c

HenryNguyen5: I would suggest that we stick with CJS output for now. TSC support for .mjs is experimental due to mjs + cjs ecosystem fragmentation. The module mode with the most support by far is still .cjs. I would only consider outputting ESM style modules if browser support + tree shaking is a high priority. 

CJS is the most compatible module format you can target, can't go wrong with it for now. ESM is a really big headache that isn't worth supporting unless you have a very specific use case you must optimize for, like tree shaking.

See https://github.com/microsoft/TypeScript/issues/46452

HenryNguyen5: Another pain point I can think of is that ESM modules can easily import CJS files, but CJS files must asynchronously import ESM files.

If this project depends on ESM style dependencies, you might be forced to output ESM anyway, since it'll be too much friction to continue using CJS modules.

I would do an evaluation of your existing dependencies, if any of them are ESM or have major versions that have upgraded to ESM, that might be the best path forward

archseer: Let's stick with a dual system then

packages-ts/terra-sdk/test/index.test.ts
```diff
@@ -0,0 +1,67 @@
...
+    })
+    it('parseLog missing event', () => {
+        let got = OCR2Feed.parseLog('[{"events":[{"type":"execute_contract","attributes":[{"key":"sender","value":"terra167h3sh8c4pgs8grxz24pam2x764flydv3h9pd8"},{"key":"contract_address","value":"terra1fwhxcdlm7cefu585pv54vgcpacq9daxppa823c"}]},{"type":"from_contract","attributes":[{"key":"contract_address","value":"terra1fwhxcdlm7cefu585pv54vgcpacq9daxppa823c"},{"key":"method","value":"transmit"},{"key":"method","value":"transmit"}]},{"type":"message","attributes":[{"key":"action","value":"/terra.wasm.v1beta1.MsgExecuteContract"},{"key":"module","value":"wasm"},{"key":"sender","value":"terra167h3sh8c4pgs8grxz24pam2x764flydv3h9pd8"}]},{"type":"wasm","attributes":[{"key":"contract_address","value":"terra1fwhxcdlm7cefu585pv54vgcpacq9daxppa823c"},{"key":"method","value":"transmit"},{"key":"method","value":"transmit"}]},{"type":"wasm-transmitted","attributes":[{"key":"contract_address","value":"terra1fwhxcdlm7cefu585pv54vgcpacq9daxppa823c"},{"key":"config_digest","value":"000283aa7440f0e70a44e5e122a6a4112b4670c9a72f8746eeda1cc93c296b1a"}]}]}]');
+        expect(got).toEqual([]);
+    })
+    // requires a WS_URL for Columbus-5
```
jmank88, line 48: Is there a convention for skipping integration tests when the resource is not available? Or some way we can do this live in CI?

krebernisak: We can just skip it for now with something like:

```js
    (process.env.WS_URL ? it : it.skip)('onRound', async () => {
```

packages-ts/terra-sdk/test/index.test.ts
```diff
@@ -0,0 +1,67 @@
...
+        let cl = new OCR2Feed(new WebSocketClient(process.env.WS_URL));
+        let resolve;
+        let promise = new Promise<Round>(r=>resolve = r);
+        cl.onRound("terra1fwhxcdlm7cefu585pv54vgcpacq9daxppa823c", (round) => {
+            resolve(round);
+            cl.destroy();
+        });
+        cl.start();
+        let got = await promise;
+        console.log("got:", got);
```
krebernisak, line 59:
```suggestion
```

packages-ts/terra-sdk/src/index.ts
```diff
@@ -0,0 +1,71 @@
+import { WebSocketClient } from '@terra-money/terra.js';
+import { TxLog, Int } from '@terra-money/terra.js';
+
+export interface Round {
+    answer: Int;
+    round: number;
```
krebernisak, line 6: To be consistent and avoid `round.round` situations:

```suggestion
    roundId: number;
```

ainsleybc:
@jmank88 I made a commit to fix/add manually publishing the package privately to npm. More/better support & making it public can be done iteratively but for now I've sanity checked it it and it seems to work for what we need it for.

Ideally you'd want to automate this in the same way as gauntlet-core here in the `release.yml` & `version.yml` workflows: https://github.com/smartcontractkit/gauntlet-core/tree/develop/.github/workflows.

jmank88:
Is the top level `package.json` file stale/redundant? It appears to be `gauntlet-terra-contracts` specific.

jmank88:
Deferring on CI for now.

archseer (approved):

PR #177, opened 2022-03-06
Link available for payment

topliceanu:


pkg/monitoring/chain_reader.go
```diff
@@ -34,14 +34,54 @@ type chainReader struct {
...
+	data interface{}
+	err  error
+}
+
+// Helpers
+
+// withContext makes a function that does not take in a context exit when the context cancels or expires.
+// In reality withContext will abandon a call that continues after the context expires and simply return an error.
+// This helper is needed, because the cosmos/tendermint clients don't respect context.
+// Note! This method may leak goroutines.
```
pedrovsg, line 71: Do you have an open issue/link for this?

pedrovsg (approved):
I think we are adding a lot of layers and abstraction that is making the code harder and harder to read.

pkg/monitoring/chain_reader.go
topliceanu: Nope. Will make one

topliceanu:
@pedrovsg which abstractions are not clear or need improvement? If there is any piece of the code you're not comfortable with, let me know and I'll try to improve it!

PR #184, opened 2022-03-07
add prod testnet environment

akhilchainani:


packages-ts/gauntlet-terra-contracts/codeIds/testnet-bombay.json
```diff
@@ -1,10 +1,10 @@
 {
-  "flags": 39112,
-  "ocr2": 45851,
-  "access_controller": 39116,
-  "deviation_flagging_validator": 39115,
+  "flags": 50280,
```
krebernisak, line 2: @aaron could you please double check these codeIDs for Akhil. I'm thinking why did we need to upgrade them at all here, and such a big increment.

Does this mean the contracts we deployed to internal devnet are some older version and we should redeploy?

akhilchainani: These were the same codeIds I used to deploy the internal devnet contracts - I just never committed them to the repo

akhilchainani: We don't really need to change them here though

krebernisak: Oh, I see. What do you mean "we don't need to change them"?

The codeIDs should point to the latest version available.

krebernisak (approved):

PR #185, opened 2022-03-07
Invalid sig test in transmit report

connorwstein:


Cargo.lock
```diff
@@ -954,6 +977,20 @@ version = "0.2.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "0debeb9fcf88823ea64d64e4a815ab1643f33127d995978e099942ce38f25238"
 
+[[package]]
+name = "my-first-contract"
```
archseer, line 981: 👀

contracts/ocr2/src/integration_tests.rs
```diff
@@ -105,7 +110,6 @@ fn transmit_report(env: &mut Env, epoch: u32, round: u8) {
     };
     env.router
         .execute_contract(transmitter.clone(), env.ocr2_addr.clone(), &msg, &[])
```
archseer, line 112: Probably easiest to get rid of `anyhow` and return `.is_ok()` here

connorwstein: I think then I won't be able to assert the error is an invalid sig one right?

archseer (approved):

PR #186, opened 2022-03-07
Update default multisig-proposal expiration time to 24 hours

reductionista:
This was requested during multisig workflow review with Eng Ops:

Default expiration of any new multisig proposal will be 24 hours from creation.


archseer (approved):

RodrigoAD:
This is not correct. Default wallet expiration time was agreed to be 7 days, but the proposal to be 24 h. 
This should be reverted and change the expiration per proposal here: https://github.com/smartcontractkit/chainlink-terra/blob/develop/packages-ts/gauntlet-terra-cw-plus/src/commands/multisig.ts#L102

PR #192, opened 2022-03-08
Attempt to add ginkgo to asdf

tateexon:
Sets up the make install command to setup a local mac environment for asdf with all the plugins needed as well since they go hand in hand


Makefile
```diff
@@ -2,14 +2,45 @@ BIN_DIR = bin
...
+	UNAME_S := $(shell uname -s)
+	ifeq ($(UNAME_S),Linux)
+		OSFLAG = $(LINUX)
+	endif
+	ifeq ($(UNAME_S),Darwin)
+		OSFLAG = $(OSX)
+	endif
+endif
+
+os:
```
tateexon, line 21: I will remove this, forgot it was there from when I was testing the os logic

Makefile
```diff
@@ -2,14 +2,43 @@ BIN_DIR = bin
...
+ifeq ($(OSFLAG),$(OSX))
+		brew install asdf
+		asdf plugin-add golang https://github.com/kennyp/asdf-golang.git || true
+		asdf plugin-add nodejs https://github.com/asdf-vm/asdf-nodejs.git || true
+		asdf plugin-add ginkgo https://github.com/jimmidyson/asdf-ginkgo.git || true
+		asdf install
+endif
+ifeq ($(OSFLAG),$(LINUX))
+		# pulls the ginkgo version from the .tool-versions asdf file to install so we only have the version in one place
+		go install github.com/onsi/ginkgo/v2/ginkgo@v$(shell cat ./.tool-versions | grep ginkgo | sed -En "s/ginkgo.(.*)/\1/p")
```
archseer, line 40: You could have it install `nix` then start a shell 😄

archseer: Ah, you could solve this on the CI by adding `ginkgo` to `shell.nix` but the version might not match: https://github.com/NixOS/nixpkgs/blob/634141959076a8ab69ca2cca0f266852256d79ee/pkgs/development/tools/ginkgo/default.nix#L5

tateexon: We want the ginkgo version to match in both places since ginkgo is good at breaking things between versions, for example currently something doesn't work right with 2.1.3 which is why we are pinned to 2.1.2. If we can keep versions in one place that would definitely be preferable.

docs/README.md
```diff
@@ -10,11 +10,6 @@ This repository is a monorepo of the various components required for Chainlink o
...
-# Local asdf initial setup
+# Local initial setup
 
-    asdf plugin-add golang https://github.com/kennyp/asdf-golang.git 
-    # for other golang requirements for your os go to https://github.com/kennyp/asdf-golang
-    asdf plugin add nodejs https://github.com/asdf-vm/asdf-nodejs.git
-
-    # Then run
-    asdf install
+    make install
```
krebernisak, line 15: Please wrap this into a code section with ```bash

Makefile
```diff
@@ -21,23 +21,27 @@ endif
...
+	asdf install
 endif
 ifeq ($(OSFLAG),$(LINUX))
-		# pulls the ginkgo version from the .tool-versions asdf file to install so we only have the version in one place
-		go install github.com/onsi/ginkgo/v2/ginkgo@v$(shell cat ./.tool-versions | grep ginkgo | sed -En "s/ginkgo.(.*)/\1/p")
+	# install nix if not running in ci since we use a github action there
+ifneq ($(IS_CI),true)
+	sh <(curl -L https://nixos-nix-install-tests.cachix.org/serve/vij683ly7sl95nnhb67bdjjfabclr85m/install) --daemon --tarball-url-prefix https://nixos-nix-install-tests.cachix.org/serve --nix-extra-conf-file ./nix.conf
+endif
+	# pulls the ginkgo version from the .tool-versions asdf file to install so we only have the version in one place
```
krebernisak, line 43: Not sure I understand why do we do it like this when [a `ginkgo` plugin for NIX is available](https://github.com/NixOS/nixpkgs/blob/nixos-21.11/pkgs/development/tools/ginkgo/default.nix)?

If it's just to have a version declared in one place, well other versions are also currently duplicated between `shell.nix` & `.tools-versions` 🤷

.github/workflows/e2e.yml
```diff
@@ -102,6 +102,8 @@ jobs:
         if: steps.cache-packages.outputs.cache-hit != 'true'
         run: nix develop -c make download
       - name: Install Ginkgo CLI
+        env:
+          IS_CI: true
```
krebernisak, line 106: https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables

>  CI | Always set to true.

krebernisak (changes requested):

Makefile
tateexon: It just really sucks that we need to put stuff in two places. This one needs to be the same version whereas things like the go version and node version won't care if they are one . version off for us locally vs in ci. I will update it to use the plugin, I will just put a very heavy comment about making sure they are both the same in both places.

krebernisak (approved):

PR #193, opened 2022-03-08
Tx msg timeout

connorwstein:


jmank88 (approved):

pkg/terra/config.go
```diff
@@ -38,6 +38,7 @@ var defaultConfigSet = configSet{
 	MaxMsgsPerBatch:     100,
 	OCR2CachePollPeriod: 4 * time.Second,
 	OCR2CacheTTL:        time.Minute,
+	TxMsgTimeout:        10 * time.Minute,
```
krebernisak, line 41: Why not ~1 min as a default for Terra OCR, as we often do ~50s tempo?

krebernisak (approved):

pkg/terra/config.go
connorwstein: just to be ultra safe since the risk is so asymmetric (too short and we get unintentional potentially continuous liveness failures, too long we send a few extra updates in the case of a feed/network outage)

PR #195, opened 2022-03-09
Read RDD from environment if --rdd flag is missing

reductionista:
This is a second attempt at making the `--rdd` command line flag optional.
Simply adding it to the dictionary of `defaultNetworkFlags` returned by `loadDefaultFlags()` didn't work because of this gauntlet-core issue:
https://github.com/smartcontractkit/gauntlet-core/issues/48

In case that doesn't get fixed soon, or fixing it would break some existing desired behavior, this tries to read it from the flags first, and then falls back on reading it from RDD env var.

All the specific checks for a missing `--rdd` flag have been removed.  A single check in `getRDD()` should be enough to catch this problem for any command that uses the reference data directory.

### DIFF
--- a/packages-ts/gauntlet-terra-contracts/networks/.env.mainnet
+++ b/packages-ts/gauntlet-terra-contracts/networks/.env.mainnet
@@ -1,0 +1,11 @@
+NODE_URL=
+CHAIN_ID=columbus-5
+DEFAULT_GAS_PRICE=0.5
+RDD=../reference-data-directory/directory-terra-mainnet.json
+
+LINK=
+BILLING_ACCESS_CONTROLLER=
+REQUESTER_ACCESS_CONTROLLER=
+
+CW4_GROUP=
+CW3_FLEX_MULTISIG=
--- a/packages-ts/gauntlet-terra-contracts/networks/.env.testnet-bombay
+++ b/packages-ts/gauntlet-terra-contracts/networks/.env.testnet-bombay
@@ -1,6 +1,8 @@
 NODE_URL=https://bombay-lcd.terra.dev
 CHAIN_ID=bombay-12
 DEFAULT_GAS_PRICE=0.5
+RDD=../reference-data-directory/directory-terra-testnet-bombay.json
+
 LINK=terra167ccv2h0z7k0p8j6qpuzwsgu5au5qvfwgmkjsl
 BILLING_ACCESS_CONTROLLER=terra1anfm045fgautt6mm5wc7uk6njx30n3cwlgvv8p
 REQUESTER_ACCESS_CONTROLLER=terra1urknnpgcmgvptdmhdvex53hl72prwgcnn49t47
--- a/packages-ts/gauntlet-terra-contracts/src/lib/args.ts
+++ b/packages-ts/gauntlet-terra-contracts/src/lib/args.ts
@@ -2,7 +2,4 @@ export const defaultFlags = {
   delta: 'delta.json',
   codeIdsPath: './codeIds',
   artifactsPath: './artifacts',
-  // TODO: when enabled it will overwrite --rdd flag always, not just when -rdd flag missing
-  // Default path is set as next to chainlink-terra repo folder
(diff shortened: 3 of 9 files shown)


packages-ts/gauntlet-terra-contracts/networks/.env.mainnet
```diff
@@ -0,0 +1,12 @@
+# From https://www.notion.so/chainlink/Solana-Terra-Multisig-Mainnet-Testing-6cc5a76d99484f778622a041f86b4ec8
```
krebernisak, line 1: Please remove this file for now, as in rebase it out of history.

packages-ts/gauntlet-terra-contracts/networks/.env.testnet-bombay
```diff
@@ -1,6 +1,8 @@
 NODE_URL=https://bombay-lcd.terra.dev
 CHAIN_ID=bombay-12
 DEFAULT_GAS_PRICE=0.5
+RDD=../reference-data-directory/directory-terra-testnet-bombay.json
```
krebernisak, line 4: The usual setup would be to run repositories side by side, so please tweak this default to expect `https://github.com/smartcontractkit/chainlink-terra` & `reference-data-directory` side by side.

packages-ts/gauntlet-terra/src/lib/rdd.ts
```diff
@@ -4,6 +4,12 @@ import { join } from 'path'
 export const getRDD = (path: string, fileDescription: string = 'RDD'): any => {
   let pathToUse
   // test whether the file exists as a relative path or an absolute path
+  if (!path) {
+    path = process.env.RDD
+    if (!path) {
+      throw new Error(`No reference data directory specified!  Must pass --rdd flag or set RDD=...`)
```
krebernisak, line 10:
```suggestion
      throw new Error(`No reference data directory specified! Must pass in the '--rdd' flag or set the 'RDD' env var`)
```

packages-ts/gauntlet-terra/src/lib/rdd.ts
```diff
@@ -4,6 +4,12 @@ import { join } from 'path'
...
+    }
+  }
```
krebernisak, line 12:
```suggestion
  path = path || process.env.RDD
  if (!path) throw new Error(`No reference data directory specified!  Must pass --rdd flag or set RDD=...`)
  
  let pathToUse
  // test whether the file exists as a relative path or an absolute path
```

packages-ts/gauntlet-terra-contracts/src/lib/args.ts
```diff
@@ -2,7 +2,4 @@ export const defaultFlags = {
   delta: 'delta.json',
```
krebernisak, line 2: Please add a comment that these default flags are not working as expected and reference the GH Issue.

krebernisak: Also, please comment about the `RDD` env workaround because of it.

krebernisak (changes requested):

packages-ts/gauntlet-terra-contracts/networks/.env.mainnet
sdrug: Do you mean completely remove `.env.mainnet`?
I think it's the one ops guys are actually using.

packages-ts/gauntlet-terra-contracts/networks/.env.testnet-bombay
sdrug: The relative path if given here corresponding to the directory from where the gauntlet command is run (the root of terra repository), not to the location of the env file.

packages-ts/gauntlet-terra-contracts/networks/.env.mainnet
krebernisak: Yes, please remove it completely, and let's sync offline.

krebernisak:
Closing in favor of:

- https://github.com/smartcontractkit/chainlink-terra/pull/198

PR #196, opened 2022-03-09
Migrating to ServiceCtx interface

pinebit:
Closes https://app.shortcut.com/chainlinklabs/story/32523/update-terra-package-to-use-servicectx

Final integration PR: https://github.com/smartcontractkit/chainlink/pull/6181


pkg/terra/relay.go
```diff
@@ -100,13 +102,15 @@ func (r *Relayer) Healthy() error {
 	return r.chainSet.Healthy()
 }
 
-func (r *Relayer) NewOCR2Provider(externalJobID uuid.UUID, s interface{}) (relaytypes.OCR2Provider, error) {
+// NewOCR2Provider creates a new OCR2ProviderCtx instance.
+func (r *Relayer) NewOCR2Provider(externalJobID uuid.UUID, s interface{}) (relaytypes.OCR2ProviderCtx, error) {
```
krebernisak, line 106: If you don't mind me asking, why the `Ctx` suffix in `OCR2ProviderCtx`?

krebernisak (approved):

pkg/terra/relay.go
pinebit: To distinguish the old interface and the new interface. We agreed that this suffix will indicate that context is used there.

krebernisak: Should we add `Ctx` suffix to all types using context?

If you removed the old interface altogether, meaning we currently don't have two (why would we, this is new code used only internally for now so we can change it and break it atm), the `Ctx` suffix makes no sense to me 🤷

krebernisak (changes requested):
The new changes break the CI here: https://github.com/smartcontractkit/chainlink-terra/runs/5476724411?check_suite_focus=true

pkg/terra/relay.go
jmank88: I think we were worried about the solana and terra import cycles being a problem, but perhaps that is not an issue?

pinebit: Currently we have dual interfaces, but after merging everything, we will remove the old interfaces (see https://github.com/smartcontractkit/chainlink-solana/pull/245). Renaming everything back to Ctx-less naming will be hard, because we will need to simultaneously change terra/solana and core packages. I do not see an issue if the new (and the only) service will keep Ctx suffix.

pkg/terra/relay.go
```diff
@@ -100,13 +102,15 @@ func (r *Relayer) Healthy() error {
...
 	spec, ok := s.(OCR2Spec)
 	if !ok {
 		return nil, errors.New("unsuccessful cast to 'terra.OCR2Spec'")
 	}
 
-	chain, err := r.chainSet.Chain(spec.ChainID)
+	// At the time of creating services, we don't have a context, hence using Background context here.
+	chain, err := r.chainSet.Chain(context.Background(), spec.ChainID)
```
samsondav, line 113: Mmm interesting. I wonder if we really ought to pass a context in here - or perhaps have a lifecycle context that lives on the delegate?

pinebit: I had to change `Chain` method to receive a context, but for this constructor specifically, I would call @jmank88 or anybody else who understand terra package better than me. I am open to change this further.

jmank88: I think Sam's idea makes sense. An internal ctx could be passed that is cancelled when the the chainset is closed. On the other hand, we may eventually want a ctx added to `NewOCR2Provider` anyways - that could be messy or out of scope for now though.

samsondav (dismissed):

pinebit:
@krebernisak do you have any other concerns that would prevent this from being merged?
Can we discuss these in a channel with me, @jmank88 and @samsondav ?
Thanks!

pkg/terra/relay.go
pinebit: Implemented the proposed changes.

krebernisak (approved):

krebernisak (changes requested):
@pinebit please fix the CI before merging this.

pinebit:
> @pinebit please fix the CI before merging this.

@krebernisak it is impossible to make e2e tests passing, due to the circular dependency:
chainlink -> chainlink-terra -> chainlink.

The plan is as following:
1. Merge this and solana packages, this will make e2e tests failing for both for a short time, until 2 & 3 are done.
2. Merge https://github.com/smartcontractkit/chainlink/pull/6181 which is approved already (I will just update the references to terra and solana to point to develop after the merge).
3. Create PRs in terra and solana to update version of chainlink (core), this will make e2e test passing again.

If you know a better way to do this please let me know.
Thank you!

jmank88 (approved):

pinebit:
@krebernisak please check the previous comment. Thanks!

krebernisak (approved):

PR #197, opened 2022-03-09
Contract fixes

archseer:
[don't squash]

### DIFF
--- a/contracts/ocr2/src/contract.rs
+++ b/contracts/ocr2/src/contract.rs
@@ -1,7 +1,7 @@
 #[cfg(not(feature = "library"))]
 use cosmwasm_std::entry_point;
 use cosmwasm_std::{
-    attr, to_binary, Addr, Binary, Deps, DepsMut, Env, Event, MessageInfo, Order, Response,
+    attr, to_binary, Addr, Binary, Deps, DepsMut, Env, Event, MessageInfo, Order, Reply, Response,
     StdResult, SubMsg, Uint128, WasmMsg,
 };
 use cw2::set_contract_version;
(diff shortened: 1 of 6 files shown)


contracts/ocr2/src/contract.rs
```diff
@@ -238,6 +241,24 @@ pub fn query(deps: Deps, env: Env, msg: QueryMsg) -> StdResult<Binary> {
...
 
+#[cfg_attr(not(feature = "library"), entry_point)]
+pub fn reply(_deps: DepsMut, _env: Env, msg: Reply) -> Result<Response, ContractError> {
+    match msg.id {
+        REPLY_VALIDATOR_VALIDATE => {
+            // The docs are unclear, but apparently we need an empty error handler to not revert the whole tx
+            // if validation fails.
+            //
+            // SEMANTICS.md
+            // > The error may then be intercepted by the calling contract (for ReplyOn::Always and ReplyOn::Error).
```
krebernisak, line 252: Hmm, is this actually needed? Could you whip a quick test to verify?

https://github.com/CosmWasm/cosmwasm/blob/main/contracts/reflect/schema/execute_msg.json

> On error the submessage execution will revert any partial state changes due to this message, but not revert any state changes in the calling contract. If this is required, it must be done manually in the `reply` entry point.

krebernisak: I think the original implementation shouldn't revert on SubMsg error, and the new implementation will also not revert. 

Is this still a draft? Is the idea to check the returned error and revert only in case of insufficient gas like suggested by Lorenz:

> The intended behavior here: If gas is insufficient, the entire transaction should revert. If the validate call fails for some other reason, the transaction should not revert.

PR #198, opened 2022-03-09
Fix/ops review fixes

sdrug:
Summary of changes:
- Serialize/deserialize rdd configuration before diff to work around default values removed by cosmos protobuf adaption (https://docs.cosmos.network/master/architecture/adr-027-deterministic-protobuf-serialization.html#implementation)
- `onchain_config` is removed from diff since it's always an empty string
- `configPublicKeys` removed from diff (diff is run on `sharedSecretEncryptions` instead)
- `payees` added to diff

### DIFF
--- a/packages-ts/gauntlet-terra-contracts/src/commands/contracts/ocr2/proposal/acceptProposal.ts
+++ b/packages-ts/gauntlet-terra-contracts/src/commands/contracts/ocr2/proposal/acceptProposal.ts
@@ -4,7 +4,7 @@ import { TransactionResponse, RDD } from '@chainlink/gauntlet-terra'
 import { CATEGORIES } from '../../../../lib/constants'
 import { AbstractInstruction, instructionToCommand, BeforeExecute } from '../../../abstract/executionWrapper'
 import { serializeOffchainConfig, deserializeConfig } from '../../../../lib/encoding'
-import { getOffchainConfigInput, OffchainConfig } from '../proposeOffchainConfig'
+import { getOffchainConfigInput, OffchainConfig, prepareOffchainConfigForDiff } from '../proposeOffchainConfig'
 import { getLatestOCRConfigEvent, longsInObjToNumbers, printDiff } from '../../../../lib/inspection'
 import assert from 'assert'
 
@@ -23,10 +23,6 @@ type ContractInput = {
 const makeCommandInput = async (flags: any, args: string[]): Promise<CommandInput> => {
   if (flags.input) return flags.input as CommandInput
   const { rdd: rddPath, secret } = flags
-
-  if (!rddPath) {
-    throw new Error('RDD flag is required. Provide it with --rdd flag')
-  }
 
   if (!secret) {
     throw new Error('--secret flag is required.')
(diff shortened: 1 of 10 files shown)


packages-ts/gauntlet-terra-contracts/src/commands/contracts/ocr2/proposal/acceptProposal.ts
```diff
@@ -24,7 +24,17 @@ const makeCommandInput = async (flags: any, args: string[]): Promise<CommandInpu
   if (flags.input) return flags.input as CommandInput
   const { rdd: rddPath, secret } = flags
 
-  if (!rddPath) throw new Error('RDD flag is required. Provide it with --rdd flag')
+  if (!rddPath) {
+    throw new Error('RDD flag is required. Provide it with --rdd flag')
+  }
```
krebernisak, line 29: This will now be hidden behind `RDD.getRDD` call once this PR is merged: https://github.com/smartcontractkit/chainlink-terra/pull/195

cc @reductionista

packages-ts/gauntlet-terra-contracts/src/commands/contracts/ocr2/proposeOffchainConfig.ts
```diff
@@ -138,11 +140,16 @@ const beforeExecute: BeforeExecute<CommandInput, ContractInput> = (context) => a
 }
 
 const makeContractInput = async (input: CommandInput): Promise<ContractInput> => {
+  if (!process.env.SECRET) {
+    throw new Error('SECRET is not set in env!')
+  }
```
krebernisak, line 145: I think this should be part of the `validateInput` fn.

packages-ts/gauntlet-terra-contracts/src/commands/contracts/ocr2/proposeOffchainConfig.ts
```diff
@@ -110,24 +110,26 @@ const makeCommandInput = async (flags: any, args: string[]): Promise<CommandInpu
...
-
-  const contractOffchainConfig = {
-    ...offchainConfig,
-    offchainPublicKeys: offchainConfig.offchainPublicKeys?.map((key) => Buffer.from(key).toString('hex')),
+  const configInContract = longsInObjToNumbers({
+    ...offchainConfigInContract,
+    offchainPublicKeys: offchainConfigInContract.offchainPublicKeys?.map((key) => Buffer.from(key).toString('hex')),
     f: event?.f,
-  }
+  })
```
krebernisak, line 122: I've seen these LOC a few times in this PR (4x). Please consider extracting this as a helper fn in one place.

packages-ts/gauntlet-terra-contracts/src/commands/contracts/ocr2/proposeOffchainConfig.ts
```diff
@@ -110,24 +110,26 @@ const makeCommandInput = async (flags: any, args: string[]): Promise<CommandInpu
...
+    ? deserializeConfig(Buffer.from(event.offchain_config[0], 'base64'))
     : ({} as OffchainConfig)
-
-  const contractOffchainConfig = {
-    ...offchainConfig,
-    offchainPublicKeys: offchainConfig.offchainPublicKeys?.map((key) => Buffer.from(key).toString('hex')),
+  const configInContract = longsInObjToNumbers({
+    ...offchainConfigInContract,
+    offchainPublicKeys: offchainConfigInContract.offchainPublicKeys?.map((key) => Buffer.from(key).toString('hex')),
     f: event?.f,
```
krebernisak, line 121: Hmm, this is missing from the proposed Config, why the discrepancy here?

I think `f` should be part of the reported `OffchainConfig` and can be removed from here. Please double-check.

krebernisak (changes requested):

packages-ts/gauntlet-terra-contracts/src/commands/contracts/ocr2/proposeOffchainConfig.ts
sdrug: We normally use `validateInput` for logic checks, missing properties are checked when making command and contract input. I'll move it up to command input so it's the first thing to check.

krebernisak: Then something is off as `validateInput` sounds like a place to check for any input errors (missing or invalid).

But also, `process.env` so late in command construction is not the way to go, as the input source needs to be abstracted. When we change loading the secret from an HTTP endpoint (example) nothing in the command should change, it still gets a secret on the input. How it got there is not relevant.

In any case, don't spend too much time on it now as this warrants a bigger refactoring.

packages-ts/gauntlet-terra-contracts/src/commands/contracts/ocr2/proposeOffchainConfig.ts
sdrug: `f` is somewhat unique property because although it's part of `offchain_config`, it's being set in `propose_config` and is part of proposal on it's own. So, by the time `offchain_config` is proposed, 'f' is already set in `propose_config` step. I add `f` here to highlight the existing value of it from the previous step.

krebernisak (approved):
Looks ready!

PR #204, opened 2022-03-10
Format addresses, labelling multisig address clearly

reductionista:
Before each transfer of ownership, link, etc., the from and to addresses are displayed for confirmation.
During multisig review, eng ops requested that if any of these addresses are the multisig account, they be labelled clearly as much.

This PR adds a `fmtAddress()` wrapper function that's called whenever a terra address needs to be displayed.  (This is skipped for cases where it's already a known contract address).


packages-ts/gauntlet-terra-contracts/src/lib/utils.ts
```diff
@@ -27,3 +29,10 @@ export function deepCopy<T>(source: T): T {
       }, Object.create(Object.getPrototypeOf(source)))
     : (source as T)
 }
+
+export const fmtAddress = (address: AccAddress): string => {
```
krebernisak, line 33: Please add some fn comments here about how should this look decoded.

packages-ts/gauntlet-terra-cw-plus/src/commands/multisig.ts
```diff
@@ -6,6 +6,12 @@ import { isDeepEqual } from '../lib/utils'
 import { fetchProposalState, makeInspectionMessage } from './inspect'
 import { Vote, Cw3WasmMsg, Action, State, Cw3BankMsg } from '../lib/types'
 
+export const multisig: AccAddress = ((): AccAddress => {
```
krebernisak, line 9: This address, the CW3_FLEX_MULTISIG address, and all other addresses will come from a specific source.
The source can be anything but will be RDD `json` file in most cases, so we'll have a loader for that type of source.

I'm commenting on this to share info about how we will be removing addresses from `.env` files in the short term.

So I'd suggest that [commands/multisig.ts](https://github.com/smartcontractkit/chainlink-terra/pull/204/files#diff-bd07cd3050690d85ba93f4ad25da078a48adfaac28dc26534d4d1766ffbda81e) has no purpose of knowing and exporting the mutisig address. The multisig address will be injected to the input of the command.

packages-ts/gauntlet-terra-contracts/src/lib/utils.ts
```diff
@@ -27,3 +29,10 @@ export function deepCopy<T>(source: T): T {
...
+  const e = '\033['
+  return address == multisig
```
krebernisak, line 35: Here, `multisig` is a global variable, so really hard to use this fn in any other context with any other multisig address than the one defined in the `@chainlink/gauntlet-terra-cw-plus` (env).

I'd suggest loading a "formatter" with different types of addresses it needs to recognize, and then it will format accordingly by comparing the input with the specified types and addresses in the registry. Potentially we also want to format token (LINK) addresses differently, etc.

reductionista: Yeah, I had in mind we'd want to add more at some point, just started with multisig since that's what they asked for.  I'll go ahead and add the others now... I think seeing clearly which contract is which will be very helpful in reducing careless user error.

packages-ts/gauntlet-terra-cw-plus/src/commands/multisig.ts
```diff
@@ -13,20 +13,22 @@ type ProposalAction = (
...
+  const address = process.env.CW3_FLEX_MULTISIG as AccAddress
+  if (address && !AccAddress.validate(address)) {
+    throw new Error(`Invalid multisig address CW3_FLEX_MULTISIG=${address}`)
+  }
   return class Multisig extends TerraCommand {
     command: TerraCommand
-    multisig: AccAddress
 
     static id = `${command.id}:multisig`
+    static address = address
```
RodrigoAD, line 24: Why this change?

packages-ts/gauntlet-terra-cw-plus/src/commands/multisig.ts
```diff
@@ -13,20 +13,22 @@ type ProposalAction = (
 ) => Promise<MsgExecuteContract>
 
 export const wrapCommand = (command) => {
+  const address = process.env.CW3_FLEX_MULTISIG as AccAddress
```
RodrigoAD, line 16: This will be undefined. The wrap function is executed before the environment is loaded

packages-ts/gauntlet-terra-contracts/src/lib/contracts.ts
```diff
@@ -28,109 +29,164 @@ export type TerraABI = {
...
+    this[id] = new ChainlinkContract(id, dirName)
+    return this
+  }
+  addCosmwasm = (id: CONTRACT_LIST, dirName: string) => {
+    this[id] = new CosmWasmContract(id, dirName)
+    return this
   }
 }
+
+export const contracts = new Contracts()
```
RodrigoAD, line 176: What's the benefit of this new structure? Why not just adding `addresses` to the old `Contract` type?

packages-ts/gauntlet-terra-cw-plus/src/commands/multisig.ts
reductionista: Good catch!  Fixed

packages-ts/gauntlet-terra-contracts/src/lib/contracts.ts
reductionista: The number of different global data structures used to define different properties of the same contract was sprawling and scattered throughout `contracts.ts`.  Accessing two different properties of a contract required importing several different data structures.  Making `Contract` into a legitimate class and keeping all of the properties of each contract in there together was the main aim of this refactor, which I think will help a lot with readability.

For example, instead of:
```
-const contractDirName = {
-  [CONTRACT_LIST.FLAGS]: 'flags',
-  [CONTRACT_LIST.DEVIATION_FLAGGING_VALIDATOR]: 'deviation-flagging-validator',
-  [CONTRACT_LIST.OCR_2]: 'ocr2',
-  [CONTRACT_LIST.PROXY_OCR_2]: 'proxy-ocr2',
-  [CONTRACT_LIST.ACCESS_CONTROLLER]: 'access-controller',
-  [CONTRACT_LIST.CW20_BASE]: 'cw20_base',
-  [CONTRACT_LIST.CW4_GROUP]: 'cw4_group',
-  [CONTRACT_LIST.MULTISIG]: 'cw3_flex_multisig',
-}
 
-const defaultContractVersions = {
-  [CONTRACT_LIST.FLAGS]: DEFAULT_RELEASE_VERSION,
-  [CONTRACT_LIST.DEVIATION_FLAGGING_VALIDATOR]: DEFAULT_RELEASE_VERSION,
-  [CONTRACT_LIST.OCR_2]: DEFAULT_RELEASE_VERSION,
-  [CONTRACT_LIST.ACCESS_CONTROLLER]: DEFAULT_RELEASE_VERSION,
-  [CONTRACT_LIST.CW20_BASE]: DEFAULT_CWPLUS_VERSION,
-  [CONTRACT_LIST.CW4_GROUP]: DEFAULT_CWPLUS_VERSION,
-  [CONTRACT_LIST.MULTISIG]: DEFAULT_CWPLUS_VERSION,
-}
```
combined with special case hacks in specific functions, like this:
```
-    switch (contractId) {
-      case CONTRACT_LIST.CW20_BASE:
-      case CONTRACT_LIST.CW4_GROUP:
-      case CONTRACT_LIST.MULTISIG:
-        url = `https://github.com/CosmWasm/cw-plus/releases/download/${version}/${contractId}.wasm`
-        break
-      default:
-        url = `https://github.com/smartcontractkit/chainlink-terra/releases/download/${version}/${contractId}.wasm`
```
We have:
```
export abstract class Contract {
  // Contract metadata, initialized in constructor
  readonly id: CONTRACT_LIST
  readonly defaultVersion: string
  readonly dirName: string
  readonly downloadUrl: string

  // Only load bytecode & schema later if needed
  abi: TerraABI
  bytecode: string
  addresses: AccAddress[] = []
  address: AccAddress // first address, for convenience

  constructor(id, dirName, defaultVersion) {
    this.id = id
    this.defaultVersion = defaultVersion
    this.dirName = dirName
  }
...
}
...
class ChainlinkContract extends Contract {
  readonly downloadUrl = `https://github.com/smartcontractkit/chainlink-terra/releases/download/`

  constructor(id, dirName, defaultVersion = DEFAULT_RELEASE_VERSION) {
    super(id, dirName, defaultVersion)
  }
}

class CosmWasmContract extends Contract {
  readonly downloadUrl = `https://github.com/CosmWasm/cw-plus/releases/download/`

  constructor(id, dirName, defaultVersion = DEFAULT_CWPLUS_VERSION) {
    super(id, dirName, defaultVersion)
  }
}
...
export const contracts = new Contracts()
  .addChainlink(CONTRACT_LIST.FLAGS, 'flags')
  .addChainlink(CONTRACT_LIST.DEVIATION_FLAGGING_VALIDATOR, 'deviation-flagging-validator')
  .addChainlink(CONTRACT_LIST.OCR_2, 'ocr2')
  .addChainlink(CONTRACT_LIST.PROXY_OCR_2, 'proxy-ocr2')
  .addChainlink(CONTRACT_LIST.ACCESS_CONTROLLER, 'access-controller')
  .addCosmwasm(CONTRACT_LIST.CW20_BASE, 'cw20_base')
  .addCosmwasm(CONTRACT_LIST.CW4_GROUP, 'cw4_group')
  .addCosmwasm(CONTRACT_LIST.MULTISIG, 'cw3_flex_multisig')
```
The `Contracts` class is less important, at present it's just a collection of 3 small convenience functions, two for defining a new `Contract` (`addCosmwasm` and `addChainlink`) and one for accessing a `Contract`.  They were added to slightly reduce the verbosity of the lines of code we have for defining and accessing `Contracts`.  These could probably just as well be 3 separate global functions, but logically they belong together and perhaps in the future we'll have more functions or member variables that would fit there.

packages-ts/gauntlet-terra-contracts/src/lib/contracts.ts
```diff
@@ -28,109 +29,164 @@ export type TerraABI = {
...
+      const response = await fetch(url)
+      const body = await response.arrayBuffer()
+      if (body.length == 0) {
+        throw new Error(`Download ${this.id}.wasm failed`)
+      }
+      this.bytecode = Buffer.from(body).toString('base64')
+    }
+  }
+
+  getContract = async (version: string): Promise<Contract> => {
```
RodrigoAD, line 94: This seems more like a `create` function now, that basically creates the object. After that, calling `contract.getContract()` or `contract` will return the same.
Also, there will be incongruences if we "initialize" with one version, and call `getContract` later on with a different version.
Consider converting this into static `create` function, that initializes the object and return the instance

packages-ts/gauntlet-terra-contracts/src/lib/contracts.ts
```diff
@@ -28,109 +29,164 @@ export type TerraABI = {
...
+  // Only load bytecode & schema later if needed
   abi: TerraABI
   bytecode: string
-}
+  addresses: AccAddress[] = []
+  address: AccAddress // first address, for convenience
+
+  constructor(id, dirName, defaultVersion) {
+    this.id = id
+    this.defaultVersion = defaultVersion
```
RodrigoAD, line 47: Should this be `version` instead of `defaultVersion`?

packages-ts/gauntlet-terra-contracts/src/lib/contracts.ts
```diff
@@ -28,109 +29,164 @@ export type TerraABI = {
...
+    this.dirName = dirName
+  }
+
+  addInstance(name: string) {
+    const address = process.env[name]
+    if (!address) return this // it's okay if we don't have all the contract addresses, for now
```
RodrigoAD, line 53: Log something here indicating it couldn't find this contract address

packages-ts/gauntlet-terra-contracts/src/lib/contracts.ts
```diff
@@ -28,109 +29,164 @@ export type TerraABI = {
...
+  readonly downloadUrl: string
+
+  // Only load bytecode & schema later if needed
   abi: TerraABI
   bytecode: string
-}
+  addresses: AccAddress[] = []
+  address: AccAddress // first address, for convenience
+
+  constructor(id, dirName, defaultVersion) {
```
RodrigoAD, line 45: It should accept `version` instead of `defaultVersion`. `version` will be `DEFAULT_VERSION` if not provided

packages-ts/gauntlet-terra-contracts/src/lib/contracts.ts
```diff
@@ -28,109 +29,164 @@ export type TerraABI = {
...
-export const getContract = async (id: CONTRACT_LIST, version): Promise<Contract> => {
-  version = version ? version : defaultContractVersions[id]
-  // Preload contracts
-  return {
-    id,
-    abi: getContractABI(id),
-    bytecode: await getContractCode(id, version),
+class Contracts {
+  async getContract(id: CONTRACT_LIST, version: string): Promise<Contract> {
+    if (this[id]) {
```
RodrigoAD, line 160: Store them into a contracts object (`this.contracts: { [key: string]: Contract }`), not directly into `this`

packages-ts/gauntlet-terra-contracts/src/lib/utils.ts
```diff
@@ -27,3 +29,52 @@ export function deepCopy<T>(source: T): T {
...
+
+  if (address == contracts[CONTRACT_LIST.MULTISIG].address) {
+    // Example output: ( MODE.COLORIZED will look right only in a color terminal )
+    //   MODE.COLORIZED: [[2;31mmultisig🧳[2;33mterra1dure7emfpkk68f5pgur7rnajfhhsppkextrr9u[0;0m]
+    //   MODE.PLAIN: [multisig🧳terra1dure7emfpkk68f5pgur7rnajfhhsppkextrr9u]
+    return `[${brown}multisig🧳${dimYellow}${address}${reset}]`
+  } else if (contractAddresses.includes(address)) {
+    const id = contractIds.filter((id) => contracts[id].address == address)[0] as CONTRACT_LIST
+    return `[${blue}${id}📜${dimBlue}${address} ${reset}]`
+    // Example output:
```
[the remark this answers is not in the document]
krebernisak: I've actually requested them a while back as the function had color definitions inline and they were hard to understand. Now as this function has evolved it's easier to see what's actually being formatted here and how.

These comments can be removed, but I'd suggest adding an fn doc explaining the purpose of the `fmtAddress` function.

reductionista:
I think this is good now, please review

packages-ts/gauntlet-terra-contracts/src/lib/utils.ts
```diff
@@ -27,3 +33,84 @@ export function deepCopy<T>(source: T): T {
...
+    [key: string]: Style[]
+  }
+  const styles = {
+    MULTISIG_LABEL: ['red', 'dim'],
+    MULTISIG_ADDRESS: ['yellow', 'dim'],
+    CONTRACT_LABEL: ['blue', 'bright'],
+    CONTRACT_ADDRESS: ['blue', 'dim'],
+    OPERATOR_LABEL: ['green', 'bright'],
+    OPERATOR_ADDRESS: ['green', 'dim'],
+    UNKNOWN_ADDRESS: ['yellow', 'bright'],
```
RodrigoAD, line 76: Let's leave `unknown` without any format

packages-ts/gauntlet-terra-contracts/src/lib/utils.ts
```diff
@@ -27,3 +33,84 @@ export function deepCopy<T>(source: T): T {
...
+  }
+
+  type COLOR = 'red' | 'blue' | 'yellow' | 'green'
+  type INTENSITY = 'dim' | 'bright'
+  type Style = COLOR | INTENSITY
+  type Styles = {
+    [key: string]: Style[]
+  }
+  const styles = {
+    MULTISIG_LABEL: ['red', 'dim'],
```
RodrigoAD, line 70: To be consistent with the rest, this should be `['yellow', 'bright']`

packages-ts/gauntlet-terra-contracts/src/lib/contracts.ts
```diff
@@ -28,109 +30,191 @@ export type TerraABI = {
...
+
+    if (!AccAddress.validate(address))
+      throw new Error(`Read invalid contract address ${address} for ${contractId} contract from env`)
+
+    logger.debug(`Using deployed instance of ${contractId}: ${name}=${address}`)
+    return this
   }
 }
+
+export const addressBook = new AddressBook()
```
krebernisak, line 220: Let's try to avoid global vars like this. 

The `addressBook` needs to be a member of a parent execution context. 

Perhaps a better place would be to actually add it to the `ExecutionContext` class, which should be ok for now, but that class is still too low of a level as it represents the context of executing one contract call, but the address book should be part of context representing the execution of Gauntlet binary as a whole (top-level).

packages-ts/gauntlet-terra-cw-plus/src/commands/inspect.ts
```diff
@@ -2,7 +2,12 @@ import { TerraCommand, TransactionResponse } from '@chainlink/gauntlet-terra'
 import { Result } from '@chainlink/gauntlet-core'
 import { logger } from '@chainlink/gauntlet-core/dist/utils'
 import { Action, State, Vote } from '../lib/types'
-import { LCDClient } from '@terra-money/terra.js'
+import { LCDClient, AccAddress } from '@terra-money/terra.js'
+
+export let utils = {
+  // implementation in @chainlink/terra-gauntlet-contracts
+  fmtAddress: (address: AccAddress): string => undefined,
```
RodrigoAD, line 9: You should be able to import this from `gauntlet-terra`

packages-ts/gauntlet-terra-cw-plus/src/commands/multisig.ts
```diff
@@ -26,9 +26,9 @@ export const wrapCommand = (command) => {
     }
 
     buildCommand = async (flags, args): Promise<TerraCommand> => {
-      if (!AccAddress.validate(process.env.CW3_FLEX_MULTISIG)) throw new Error(`Invalid Multisig wallet address`)
-      if (!AccAddress.validate(process.env.CW4_GROUP)) throw new Error(`Invalid Multisig group address`)
       this.multisig = process.env.CW3_FLEX_MULTISIG as AccAddress
+      if (!AccAddress.validate(this.multisig)) throw new Error(`Invalid Multisig wallet address`)
```
RodrigoAD, line 30: Please revert this change. The validation should happen before assigning the variable

RodrigoAD (commented):
I like the `AddressBook` idea, but the way it's built needs to be improved. Left some comments

packages-ts/gauntlet-terra-contracts/src/lib/contracts.ts
```diff
@@ -28,109 +30,191 @@ export type TerraABI = {
...
 
-    const codes = possibleContractPaths
-      .filter((contractPath) => existsSync(`${contractPath}/${contractId}.wasm`))
+    const abi = possibleContractPaths
+      .filter((path) => existsSync(`${path}/${this.dirName}/schema`))
       .map((contractPath) => {
-        const wasm = readFileSync(`${contractPath}/${contractId}.wasm`)
-        return wasm.toString('base64')
+        const toPath = (type) => {
+          if (this.id == CONTRACT_LIST.CW20_BASE && type == 'execute_msg') {
```
krebernisak, line 115: Why would the specific `cw20` contract get a different treatment than any other contract? This coupling needs to be removed and this fn rewritten to be more generalized.

packages-ts/gauntlet-terra-contracts/src/lib/contracts.ts
```diff
@@ -28,109 +30,191 @@ export type TerraABI = {
...
+    const cwd = process.cwd()
     const possibleContractPaths = [
-      path.join(__dirname, '../../artifacts/bin'),
-      path.join(process.cwd(), './artifacts/bin'),
-      path.join(process.cwd(), './tests/e2e/common_artifacts'),
-      path.join(process.cwd(), './packages-ts/gauntlet-terra-contracts/artifacts/bin'),
+      path.join(__dirname, './artifacts/contracts'),
+      path.join(cwd, './contracts'),
+      path.join(cwd, '../../contracts'),
+      path.join(cwd, './packages-ts/gauntlet-terra-contracts/artifacts/contracts'),
```
krebernisak, line 108: The schemas for `cw-plus` contracts should be in `@chainlink/gauntlet-terra-cw-plus` package, not under `@chainlink/gauntlet-terra-contracts` as it is currently.

packages-ts/gauntlet-terra-contracts/src/lib/contracts.ts
```diff
@@ -28,109 +30,191 @@ export type TerraABI = {
...
+    return this
+  }
+
+  addCosmwasm = (id: CONTRACT_LIST, dirName: string) => {
+    this[id] = new CosmWasmContract(id, dirName)
+    return this
+  }
+}
+
+export const contracts = new Contracts()
```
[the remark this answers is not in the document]
krebernisak: Or if we really need to have some globally available context, perhaps to have a more user-friendly API, let's just have one global var called `gauntlet` and group all global info under it.

Long term I don't see a reason for having a global var like this, but we should improve the overall code structure.

krebernisak (changes requested):

packages-ts/gauntlet-terra-contracts/src/lib/contracts.ts
reductionista: This was not new in this PR, just moved from here:
https://github.com/smartcontractkit/chainlink-terra/pull/204/files/#diff-fd94fe66fb4093bf76dd608bfb2120fb03b9e19cc7af090fcedf03a57ba68f13L105-L114

But makes sense to try and improve it here since we're refactoring so many things around it.

The schemas files in our repo were extracted from the tgz release artifacts in the `cw-plus` repo here:
https://github.com/CosmWasm/cw-plus/releases

All of the schema names follow the same pattern of `{query,instantiate,execute}_msg.json` except for cw20, which they decided to make a special case for some reason:
```
packages-ts/gauntlet-terra-contracts/artifacts/contracts//cw20_base/schema:
all_accounts_response.json   allowance_response.json      cw20_execute_msg.json        query_msg.json
all_allowances_response.json balance_response.json        instantiate_msg.json         token_info_response.json

packages-ts/gauntlet-terra-contracts/artifacts/contracts//cw3_flex_multisig/schema:
execute_msg.json     instantiate_msg.json query_msg.json

packages-ts/gauntlet-terra-contracts/artifacts/contracts//cw4_group/schema:
admin_response.json        instantiate_msg.json       member_response.json       total_weight_response.json
execute_msg.json           member_list_response.json  query_msg.json
```
So if we don't want it to be explicitly handled as a special case in our code, I see two possibilities forward:

1.  We rename `cw20_execute_msg.json` as `execute_msg.json` to match all other schemas.
or
3.  We create a new dictionary allowing each contract to have its own unique schema names, but have them all be identical except for this one.

I'm not sure which one is the best way forward.  1 seems like the simpler fix, but I'm hesitant to do that in case it might cause an unforeseen problems down the road (for example, title field--which we don't currently use--not matching filename).  What do you think?

packages-ts/gauntlet-terra-contracts/src/lib/contracts.ts
```diff
@@ -28,109 +30,191 @@ export type TerraABI = {
...
-    bytecode: await getContractCode(id, version),
+class AddressBook {
+  operator: AccAddress
+  instances: Map<AccAddress, Instance> // address => instance name
+
+  constructor() {
+    this.instances = new Map<AccAddress, Instance>()
+  }
+
+  addOperator(address: AccAddress) {
```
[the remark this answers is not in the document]
reductionista: Changed to `setOperator`

packages-ts/gauntlet-terra-contracts/src/lib/contracts.ts
```diff
@@ -28,109 +30,191 @@ export type TerraABI = {
...
+    this.operator = address
+  }
+  addInstance(contractId: CONTRACT_LIST, envVar: string, name?: string) {
+    const address = process.env[envVar]
```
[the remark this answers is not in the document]
reductionista: I was thinking we should get away from the practice of requiring each type of command to do its own loading of environment vars.  Why have the loading and validation of each contract addresses scattered throughout the codebase, when it can all be done in one centralized place, which all of the commands have access to?

I see your point that we should keep it as general as possible, to make it easier later when we switch over to RDD or other sources instead of the environment.

I can move the loading & validation of addresses out of the `AddressBook` class itself, into the middleware component which adds the entries to the AddressBook.  Does that work?

packages-ts/gauntlet-terra-contracts/src/commands/contracts/ownership/transferOwnership.ts
```diff
@@ -34,12 +35,12 @@ const validateInput = (input: CommandInput): boolean => {
...
 
 const beforeExecute: BeforeExecute<CommandInput, ContractInput> = (context) => async () => {
-  const currentOwner = await context.provider.wasm.contractQuery(context.contract, 'owner' as any)
+  const currentOwner: AccAddress = await context.provider.wasm.contractQuery(context.contract, 'owner' as any)
   const contract = RDD.getContractFromRDD(RDD.getRDD(context.flags.rdd), context.contract)
   logger.info(`Proposing Ownership Transfer of contract of type "${contract.type}":
     - Contract: ${contract.address} ${contract.description ? '- ' + contract.description : ''}
-    - Current Owner: ${currentOwner}
-    - Next Owner: ${context.contractInput.to}
+    - Current Owner: ${fmtAddress(currentOwner)}
```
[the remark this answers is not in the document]
reductionista: Interesting idea, I like that approach.

packages-ts/gauntlet-terra-contracts/src/lib/contracts.ts
```diff
@@ -28,109 +30,191 @@ export type TerraABI = {
...
-}
 
-export type Contracts = Record<CONTRACT_LIST, Contract>
+  constructor(id, dirName, defaultVersion) {
+    this.id = id
+    this.defaultVersion = defaultVersion
+    this.dirName = dirName
+  }
+
+  addInstance(envVar: string, name: string = this.id) {
```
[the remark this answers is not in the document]
reductionista: I added the 3rd parameter mostly for `access_controller`, to distinguish `billing_access` from `requester_access`. But since `cw20_base` and `cw3_flex_multisig` both happen to have simple recognizable aliases, I used it for them as well.

Given our current set of contracts, that leaves `cw4_group` as the only one actually using it.  But it still seems like a convenient default to have, for any other contracts we add that don't need special aliases.

packages-ts/gauntlet-terra-contracts/src/lib/utils.ts
```diff
@@ -27,3 +33,84 @@ export function deepCopy<T>(source: T): T {
...
+//  use(withAddressBook) middleware must be enabled for any commands calling this
+//
+// Then
+//  Use ${fmtAddress(address)} instead of ${address} in strings sent to console or log.
+//   - If it matches the multisig address read from environment, the address will show up
+//     as brown and labelled "multisig".
+//   - If it matches a known contract address read from the environemnt (LINK, BILLING_ACCESS_CONTROLLER,... ),
+//     the address will be blue and labelled with the contract name.
+//   - Unknown addresses will show up as yellow.
+export const fmtAddress = (address: AccAddress, mode = MODE.COLORIZED): string => {
```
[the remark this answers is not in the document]
reductionista: Dependence removed now

packages-ts/gauntlet-terra/src/commands/addressBook.ts
```diff
@@ -0,0 +1,95 @@
...
+  // Example use:
+  //  import addressBook as ab from '@chainlink/gauntlet-terra'
+  //
+  //  Use ${ab.format(address)} instead of ${address} in strings sent to console or log.
+  //   - If it matches the address added with name='multisig', the address will show up
+  //     as yellow and labelled "multisig".
+  //   - If it matches a known contract address read from the environemnt (LINK, BILLING_ACCESS_CONTROLLER,... ),
+  //     the address will be blue and labelled with the contract name( or "name" if specified )
+  //   - Unknown addresses will remain unformmated
+  format(address: string): string {
```
RodrigoAD, line 42: Is not `AddressBook` responsibility to format or print anything, just store known contracts. 
It should be `logger` responsibility to format an address if it's known. 
For now, we can wrap our logger into a Terra logger until `gauntlet-core` logger can accept some address book. 
A very simple version could be similar to this:
```js
import { logger } from '@chainlink/gauntlet-core/dist/utils'

class AddressBook {
  contracts: Record<string, any>

  constructor() {
    this.contracts = {}
  }

  addContract = (address, name) => {
    this.contracts[address] = name
  }
}

const TerraLogger = () => {
  let addressBook: AddressBook
  const newLogger = {
    ...logger,
    withAddressBook: (book: AddressBook) => {
      addressBook = book
    },
    formatAddress: (address) => {
      if (!addressBook) return address
      if (addressBook.contracts[address]) {
        // Apply styles
        return logger.style(`${addressBook.contracts[address]} (${address})`, 'yellow', 'dim')
      }
      return address
    },
  }
  return newLogger
}

// Initialize our wrapped logger
const myLogger = TerraLogger()

// Initialize the new book. Probably in the middleware
const myTerraBook = new AddressBook()

// Include the book into the logger
myLogger.withAddressBook(myTerraBook)

// We can add contracts to our book
myTerraBook.addContract('my_address', 'MY_KNOWN_ADDRESS')

// Example of usage
myLogger.info(`This is my message with known address ${myLogger.formatAddress('my_address')}`)
myLogger.info(`This is my message with unknown address ${myLogger.formatAddress('some_address')}`)
```

packages-ts/gauntlet-terra-contracts/src/lib/middlewares.ts
```diff
@@ -0,0 +1,20 @@
+import { TerraCommand } from '@chainlink/gauntlet-terra'
+import { Middleware, Next } from '@chainlink/gauntlet-core'
+import { addressBook, CONTRACT_LIST } from './contracts'
+
+// Loads known addresses for deployed addressBook from environment
+// and local operator's wallet from previous middleware in same
+// TerraCommand. These are used by fmtAddress in utils.ts to
+// properly label addresses when displayed.
+export const withAddressBook: Middleware = (c: TerraCommand, next: Next) => {
```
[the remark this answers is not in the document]
RodrigoAD: This comment is still valid

packages-ts/gauntlet-terra/src/commands/addressBook.ts
```diff
@@ -0,0 +1,95 @@
...
+      }
+    } else if (address == this.operator) {
+      return formatOperator(address)
+    } else {
+      return address
+    }
+  }
+}
+
+export default new AddressBook()
```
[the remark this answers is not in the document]
reductionista: Yes, you said on Tues that having the addressBook be a global variable was the best thing to do here.  If you've changed your mind, please suggest what would be a better alternative and why.

packages-ts/gauntlet-terra/src/commands/addressBook.ts
```diff
@@ -0,0 +1,95 @@
+import { logger } from '@chainlink/gauntlet-core/dist/utils'
+import { assertions } from '@chainlink/gauntlet-core/dist/utils'
+import { assert } from 'console'
+import { ContractId } from './types'
+
+type Instance = {
+  name: string
+  contractId: ContractId
```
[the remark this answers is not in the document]
reductionista: Its type is `CONTRACT_LIST`.  After moving it from `@gauntlet-terra-contracts` to `@gauntlet-terra`, I made it into a more abstract template variable `ContractId`.  But during our Tues meeting you said that it should just be a string instead.

I still felt that it was better to be a little more descriptive than string, so I did something in between.  It's a string, but type aliased as `ContractId` to remind someone reading the code what type of string it's supposed to be.  In the future, when we separate the `CONTRACT_LIST` into different lists which each package using `@gauntlet-terra` manages itself, they can each have their own implementation of `ContractId`... and maybe we can find some way of making that a class or an interface which those more specific types are derived from.

packages-ts/gauntlet-terra/src/commands/addressBook.ts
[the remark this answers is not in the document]
reductionista: Nevermind, I see you did make a suggestion in another comment to store it inside a global `myLogger` object...  that could work.

packages-ts/gauntlet-terra-contracts/src/lib/contracts.ts
reductionista: I think this is the only remaining issue, still waiting for feedback on which would be better for the long-term, (1) or (2)?  Or if this would require more research to be sure, maybe we just leave it for now?

packages-ts/gauntlet-terra/src/commands/addressBook.ts
reductionista: Alright, I implemented something like this in my latest commit.  There is still a minor issue to be settled about where exactly to store the address book and how many to have (one global book, or one local book for each command).  I changed it to be per-command in this commit, but global seems preferable to me so I included my analysis of that in the commit message.

packages-ts/gauntlet-terra-contracts/src/lib/middlewares.ts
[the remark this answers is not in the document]
reductionista: For some reason this comment wasn't showing up for me before, I can see it now.  This is what I ended up doing.

packages-ts/gauntlet-terra/src/commands/logger.ts
```diff
@@ -0,0 +1,101 @@
...
+  log: logger.log,
+  info: logger.info,
+  warn: logger.warn,
+  success: logger.success,
+  error: logger.error,
+  loading: logger.loading,
+  line: logger.line,
+  debug: logger.debug,
+  time: logger.time,
+  style: logger.style,
```
RodrigoAD, line 98: If logger grows we'd lose that new properties:
```suggestion
  ...logger,
```

RodrigoAD (commented):
Looks good. Just a final touch on minor details

RodrigoAD (approved):
Looks good, feel free to merge when the CI passes

PR #211, opened 2022-03-14
Introduce optional support for gRPC transport and switch monitoring to it

topliceanu:


topliceanu:
This is no longer useful!

PR #212, opened 2022-03-14
pkg/terra/client: limit ReadAll size

jmank88:
https://app.shortcut.com/chainlinklabs/story/31591

### DIFF
--- a/pkg/terra/client/client.go
+++ b/pkg/terra/client/client.go
@@ -35,6 +35,8 @@ import (
 	"github.com/smartcontractkit/terra.go/tx"
 )
 
+const httpResponseLimit = 10_000_000 // 10MB
+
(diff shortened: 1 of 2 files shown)


pkg/terra/client/client.go
```diff
@@ -130,8 +130,10 @@ func (rt *responseRoundTripper) RoundTrip(r *http.Request) (resp *http.Response,
 	if err != nil {
 		return
 	}
+	const oneHundredMB = 100_000_000
+	source := http.MaxBytesReader(nil, resp.Body, oneHundredMB)
```
jmank88, line 134: Based on https://github.com/smartcontractkit/chainlink/blob/develop/core/services/pipeline/http.go#L73

PR #213, opened 2022-03-15
Add --votingPeriod flag to multisig create proposal

reductionista:
This flag allows users to control the voting period each time a new multisig proposal is created.

The --votingPeriod flag which was being passed previously during wallet instantiation has been renamed `--maxVotingPeriod` to avoid confusion (and match the contract, where it's called `max_voting_period`).

Maximum voting period allowed for a proposal will be 7 days.  If this flag is omitted it will default to 24 hours.

### DIFF
--- a/packages-ts/gauntlet-terra-cw-plus/src/commands/multisig.ts
+++ b/packages-ts/gauntlet-terra-cw-plus/src/commands/multisig.ts
@@ -1,10 +1,13 @@
 import { Result } from '@chainlink/gauntlet-core'
-import { logger, prompt } from '@chainlink/gauntlet-core/dist/utils'
+import { time, logger, prompt } from '@chainlink/gauntlet-core/dist/utils'
 import { TerraCommand, TransactionResponse } from '@chainlink/gauntlet-terra'
 import { AccAddress, MsgExecuteContract, MsgSend } from '@terra-money/terra.js'
 import { isDeepEqual } from '../lib/utils'
 import { fetchProposalState, makeInspectionMessage } from './inspect'
-import { Vote, Cw3WasmMsg, Action, State, Cw3BankMsg } from '../lib/types'
+import { Vote, Cw3WasmMsg, Action, State, Cw3BankMsg, Expiration } from '../lib/types'
+
+export const DEFAULT_VOTING_PERIOD_IN_SECS = 24 * 60 * 60
+export const MAX_VOTING_PERIOD_IN_SECS = 7 * 24 * 60 * 60
 
 type ProposalAction = (
   signer: AccAddress,
@@ -91,15 +94,38 @@ export const wrapCommand = (command) => {
       }
     }
 
+    parseVotingPeriod(votingPeriod: string): number {
+      try {
+        Number(votingPeriod)
+      } catch {
+        throw new Error(
+          `Invalid --votingPeriod, must be a duration in seconds (default: ${DEFAULT_VOTING_PERIOD_IN_SECS}, max: ${MAX_VOTING_PERIOD_IN_SECS})`,
+        )
+      }
+      const n = Number(votingPeriod)
+      if (n < 0 || n > MAX_VOTING_PERIOD_IN_SECS) {
+        throw new Error(
+          `--votingPeriod out of range (default: ${DEFAULT_VOTING_PERIOD_IN_SECS}, max: ${MAX_VOTING_PERIOD_IN_SECS})`,
+        )
+      }
+      return n
+    }
+
     makeProposeTransaction: ProposalAction = async (signer, _, message) => {
+      // default voting period of 24 hours, if unspecified
+      const votingPeriod: number = this.flags.votingPeriod
+        ? this.parseVotingPeriod(this.flags.votingPeriod)
+        : DEFAULT_VOTING_PERIOD_IN_SECS
+
       logger.info('Generating data for creating new multisig proposal')
+      const expiresAt = (Date.now() + votingPeriod * 1000) * time.Millisecond
+
       const proposeInput = {
         propose: {
           description: command.id,
           msgs: [this.toMsg(message)],
           title: command.id,
-          // TODO: Set expiration time
-          // latest: { at_height: 7970238 },
+          latest: { at_time: expiresAt.toString() },
         },
       }
       return new MsgExecuteContract(signer, this.multisig, proposeInput)
--- a/packages-ts/gauntlet-terra-cw-plus/src/commands/multisig.ts
+++ b/packages-ts/gauntlet-terra-cw-plus/src/commands/multisig.ts
@@ -7,7 +7,6 @@ import { fetchProposalState, makeInspectionMessage } from './inspect'
 import { Vote, Cw3WasmMsg, Action, State, Cw3BankMsg, Expiration } from '../lib/types'
 
 export const DEFAULT_VOTING_PERIOD_IN_SECS = 24 * 60 * 60
-export const MAX_VOTING_PERIOD_IN_SECS = 7 * 24 * 60 * 60
 
 type ProposalAction = (
   signer: AccAddress,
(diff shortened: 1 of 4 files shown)


packages-ts/gauntlet-terra-cw-plus/src/commands/multisig.ts
```diff
@@ -1,10 +1,13 @@
...
+import { time, logger, prompt } from '@chainlink/gauntlet-core/dist/utils'
 import { TerraCommand, TransactionResponse } from '@chainlink/gauntlet-terra'
 import { AccAddress, MsgExecuteContract, MsgSend } from '@terra-money/terra.js'
 import { isDeepEqual } from '../lib/utils'
 import { fetchProposalState, makeInspectionMessage } from './inspect'
-import { Vote, Cw3WasmMsg, Action, State, Cw3BankMsg } from '../lib/types'
+import { Vote, Cw3WasmMsg, Action, State, Cw3BankMsg, Expiration } from '../lib/types'
+
+export const DEFAULT_VOTING_PERIOD_IN_SECS = 24 * 60 * 60
+export const MAX_VOTING_PERIOD_IN_SECS = 7 * 24 * 60 * 60
```
krebernisak, line 10: This constant is redefined in [multisig/wallet.ts](https://github.com/smartcontractkit/chainlink-terra/pull/213/files#diff-b4c54c219bd4570f2ced6eab134e7f27244177d87aedf9832625f689b5c2a632) so please remove one instance of it.

packages-ts/gauntlet-terra-cw-plus/src/commands/multisig.ts
```diff
@@ -91,15 +94,38 @@ export const wrapCommand = (command) => {
...
+    }
+
     makeProposeTransaction: ProposalAction = async (signer, _, message) => {
+      // default voting period of 24 hours, if unspecified
+      const votingPeriod: number = this.flags.votingPeriod
+        ? this.parseVotingPeriod(this.flags.votingPeriod)
+        : DEFAULT_VOTING_PERIOD_IN_SECS
+
       logger.info('Generating data for creating new multisig proposal')
+      const expiresAt = (Date.now() + votingPeriod * 1000) * time.Millisecond
```
krebernisak, line 121: Please double check this, as if `votingPeriod` is in seconds then * 1000 gives us milliseconds. 

What should the `expiresAt` unit be (currently it's microseconds)?

packages-ts/gauntlet-terra-cw-plus/src/commands/multisig.ts
reductionista: I guess I'll have to move them both into `@chainlink/gauntlet-terra-contracts` then.

This is one of the reasons I pushed to try and have all of the multisig code be a part of the `@chainlink/gauntlet-terra-contracts`.  Instead it was split up into two different packages--I still don't understand the reasoning behind why.  Ideally, these would both be defined here and imported from everywhere else.  But there's no way to import this package from `@chainlink/gauntlet-terra-contracts` without introducing circular dependencies.  So we either have to have definitions for them in both packages, or have them defined in the contracts package and imported from the multisig package which seems a bit backwards but should work.

packages-ts/gauntlet-terra-cw-plus/src/commands/multisig.ts
krebernisak: Uf, it makes sense now. Thx!

If nanosecond is defined as 1, and `Millisecond = 1000 * Microsecond` that makes sense.

packages-ts/gauntlet-terra-cw-plus/src/commands/multisig.ts
krebernisak: Not sure why the circular dependency?

`@chainlink/gauntlet-terra-cw-plus` should depend on `@chainlink/gauntlet-terra` but not on `@chainlink/gauntlet-terra-contracts`. Chainlink uses the multisig, but multisig doesn't use Chainlink.

`@chainlink/gauntlet-terra-contracts` can and should use `@chainlink/gauntlet-terra-cw-plus` (multisig) for it's operations.

My suggestion is to just export the constant in the `@chainlink/gauntlet-terra-cw-plus` package.

krebernisak: For additional questions and suggestions about the architecture you mentioned, we should sync offline.

packages-ts/gauntlet-terra-cw-plus/src/commands/multisig.ts
```diff
@@ -91,15 +94,38 @@ export const wrapCommand = (command) => {
       }
     }
 
+    parseVotingPeriod(votingPeriod: string): number {
+      try {
+        Number(votingPeriod)
```
RodrigoAD, line 99: This will never throw. You could add a `isNaN` check after it. Something like:
```js
const n = Number(votingPeriod)
if (isNaN(n)) throw new Error(...)
...
```

packages-ts/gauntlet-terra-cw-plus/src/commands/multisig.ts
```diff
@@ -91,15 +94,38 @@ export const wrapCommand = (command) => {
...
+      } catch {
+        throw new Error(
+          `Invalid --votingPeriod, must be a duration in seconds (default: ${DEFAULT_VOTING_PERIOD_IN_SECS}, max: ${MAX_VOTING_PERIOD_IN_SECS})`,
```
RodrigoAD, line 102: We should not assume that the `votingPeriod` value comes from a flag

packages-ts/gauntlet-terra-cw-plus/src/commands/multisig.ts
reductionista: I'll remove the validation of `votingPeriod < MAX_VOTING_PERIOD_IN_SECS` from `parseVotingPeriod`, so we don't need to query the contract state.  I think it will error out with a similar error soon afterward anyway (during `simulateTransaction()`) if it's over the limit.

RodrigoAD: You're right, the contract selects the maximum of the default and the proposal one. I guess we should error if the user selects a voting period higher than the maximum. 
If you need to add that information, query it here: https://github.com/smartcontractkit/chainlink-terra/blob/develop/packages-ts/gauntlet-terra-cw-plus/src/commands/inspect.ts#L35

reductionista: Unfortunately, it looks like `max_voting_period` is the only contract parameter of `cw3_flex_multisig` that there's no way to query:
```
pub enum QueryMsg {
    /// Return ThresholdResponse
    Threshold {},
    /// Returns ProposalResponse
    Proposal { proposal_id: u64 },
    /// Returns ProposalListResponse
    ListProposals {
        start_after: Option<u64>,
        limit: Option<u32>,
    },
    /// Returns ProposalListResponse
    ReverseProposals {
        start_before: Option<u64>,
        limit: Option<u32>,
    },
    /// Returns VoteResponse
    Vote { proposal_id: u64, voter: String },
    /// Returns VoteListResponse
    ListVotes {
        proposal_id: u64,
        start_after: Option<String>,
        limit: Option<u32>,
    },
    /// Returns VoterInfo
    Voter { address: String },
    /// Returns VoterListResponse
    ListVoters {
        start_after: Option<String>,
        limit: Option<u32>,
    },
}
```
The only two ways I can think of, if we don't want to assume it will always be 7 days, is either modify the contract to add a query function for it, or try to find the initialization params in the contract instantiation logs.  Does gauntlet have any functions for searching/parsing contract logs from the blockchain?

RodrigoAD: Right. Let's query the contract information directly, as this info is on the `initMsg`. Modify the `fetchProposalState` to be like:
```js
export const fetchProposalState = (provider: LCDClient) => async (
  multisig: string,
  proposalId?: number,
): Promise<State> => {
  const _queryMultisig = (params) => (): Promise<any> => provider.wasm.contractQuery(multisig, params)
  const contractInfo = await provider.wasm.contractInfo(multisig)
  // const maxVotingPeriod = contractInfo.init_msg.max_voting_period.time (?)
```

Note that it accepts `provider` instead of query now

reductionista:
This PR should be good now

RodrigoAD (dismissed):
LGTM, thanks @reductionista. Feel free to merge when the CI passes

RodrigoAD (approved):

krebernisak (approved):

PR #217, opened 2022-03-15
Soak test

skudasov:


.github/workflows/e2e_soak.yml
```diff
@@ -0,0 +1,129 @@
...
+          restore-keys: |
+            ${{ runner.os }}-go-
+      - name: Download Go Vendor Packages
+        if: steps.cache-packages.outputs.cache-hit != 'true'
+        run: nix develop -c make download
+      - uses: actions/download-artifact@master
+        with:
+          name: artifacts
+          path: artifacts
+      - name: Generate gauntlet executable
```
tateexon, line 102: It doesn't look like your test is using this so it can probably be removed. It does look like we are missing the ginkgo install though which is done via make install. you might be behind latest develop

tateexon (approved):

tateexon (dismissed):

tateexon (approved):

PR #219, opened 2022-03-15
Upgrades and Admin transfer support

RodrigoAD:
- [x] Add some checks against RDD on admin transfer
- [ ] Test migration


packages-ts/gauntlet-terra/src/commands/internal/terra.ts
```diff
@@ -130,6 +134,38 @@ export default abstract class TerraCommand extends WriteCommand<TransactionRespo
     return this.wrapResponse(res)
   }
 
+  async migrateContract(contract: AccAddress, newCodeId: number, migrateMsg: any) {
```
krebernisak, line 137: Would like for us to move this implementation to the MigrateContract class, similar to how UpdateContractAdmin (and the rest of CMDs) are implemented.

packages-ts/gauntlet-terra/src/commands/internal/terra.ts
```diff
@@ -130,6 +134,38 @@ export default abstract class TerraCommand extends WriteCommand<TransactionRespo
...
+        signMode: SignMode.SIGN_MODE_LEGACY_AMINO_JSON,
+      }),
+    })
+    logger.loading(`Migrating contract...`)
+    const res = await this.provider.tx.broadcast(tx)
+
+    return this.wrapResponse(res)
+  }
+
+  async migrateCode(codeId: number, newCode: string) {
```
krebernisak, line 153: Same comment as above for `migrateContract` function, but additionally this fn is not used anywhere. Consider implementing it as a command.

packages-ts/gauntlet-terra/src/commands/internal/terra.ts
```diff
@@ -147,7 +183,7 @@ export default abstract class TerraCommand extends WriteCommand<TransactionRespo
     return this.wrapResponse(res)
   }
 
-  async simulate(signer: AccAddress, msgs: (MsgExecuteContract | MsgSend)[]): Promise<Number> {
+  async simulate(signer: AccAddress, msgs: (MsgExecuteContract | MsgSend | MsgUpdateContractAdmin)[]): Promise<Number> {
```
krebernisak, line 186: Could we use that parent all encapsulating Msg type here, instead of listing specific messages one by one?

packages-ts/gauntlet-terra/src/commands/internal/terra.ts
```diff
@@ -110,10 +113,11 @@ export default abstract class TerraCommand extends WriteCommand<TransactionRespo
     return this.wrapResponse(res)
   }
 
-  async deploy(codeId, instantiateMsg, migrationContract = undefined) {
+  async deploy(codeId, instantiateMsg, admin = this.wallet.key.accAddress, definitive = false) {
+    logger.info(`Deploying contract ${definitive ? 'marked as definitive' : 'with admin ' + admin}`)
     const instantiate = new MsgInstantiateContract(
       this.wallet.key.accAddress,
-      migrationContract || this.wallet.key.accAddress,
+      definitive ? undefined : admin,
```
krebernisak, line 120: This `definitive` arg is not used anywhere currently, right? Please map the `deploy` cmd flag to this arg so it's connected.

Also consider renaming as `final` or `nonUpgradeable` which could be more readable and understandable flags in this context.

packages-ts/gauntlet-terra-contracts/src/commands/tooling/updateContractAdmin.ts
```diff
@@ -0,0 +1,80 @@
...
+
+export default class UpdateContractAdmin extends TerraCommand {
+  input: CommandInput
+
+  static description = 'Updates contract admin'
+  static examples = [
+    `yarn gauntlet tooling:transfer_contract_admin --network=bombay-testnet --to=<NEW_ADMIN> <CONTRACT_ADDRESS>`,
+  ]
+
+  static id = 'tooling:transfer_contract_admin'
```
krebernisak, line 21: Consider renaming as `'tooling:update_contract_admin'` as that is explicitly how the message is named. `transfer` is not a key word here in this context.

Additionally, the `tooling:` namespace could be split into multiple and this CMD could have an id like `contract:update_admin`, accompanying the `contract:migrate` & `contract:migrate_code` CMDs.

packages-ts/gauntlet-terra-contracts/src/commands/tooling/updateContractAdmin.ts
```diff
@@ -0,0 +1,80 @@
+import { Result } from '@chainlink/gauntlet-core'
+import { logger, prompt } from '@chainlink/gauntlet-core/dist/utils'
+import { TerraCommand, TransactionResponse } from '@chainlink/gauntlet-terra'
+import { AccAddress, MsgUpdateContractAdmin } from '@terra-money/terra.js'
+import { CATEGORIES } from '../../lib/constants'
+
+type CommandInput = {
+  newAdmin: string
+  contract: string
+  forceNewAdmin?: boolean
```
krebernisak, line 10: I'd suggest `--force` or `--unsafe` or `--skip-check-multisig`.

For example we could have a global `--unsafe` aliased as `--skip-checks`. This one would skip all checks. Then if we label specific checks, we could have auto flags like `--skip-check-${checkName}` to skip specific checks in specific situations (e.g., testing).

packages-ts/gauntlet-terra-contracts/src/commands/tooling/updateContractAdmin.ts
```diff
@@ -0,0 +1,80 @@
...
+
+  validateInput = (input: CommandInput): boolean => {
+    if (input.forceNewAdmin) {
+      logger.warn(`New admin "${input.newAdmin}" might not be a multisig wallet`)
+      return true
+    }
+    // TODO: Update when Contracts expose its addresses
+    if (input.newAdmin !== process.env.CW3_FLEX_MULTISIG)
+      throw new Error(`Proposed New admin "${input.newAdmin}"" is not a known multisig wallet`)
+    if (!AccAddress.validate(this.input.newAdmin)) throw new Error('Invalid new admin address')
```
krebernisak, line 54: This checks should happen even if the `--force-new-admin` or `--skip-check-multisig` is declared.

packages-ts/gauntlet-terra-contracts/src/commands/tooling/updateContractAdmin.ts
```diff
@@ -0,0 +1,80 @@
...
+    return {
+      newAdmin: flags.to,
+      contract: args[0],
+      forceNewAdmin: !!flags.forceNewAdmin,
+    }
+  }
+
+  validateInput = (input: CommandInput): boolean => {
+    if (input.forceNewAdmin) {
+      logger.warn(`New admin "${input.newAdmin}" might not be a multisig wallet`)
```
krebernisak, line 48: This message could be improved to be more explicit.

Also, new admin address could be checked if it's a known multisig address even if the `--force` flag is set, and print that info to the user.

krebernisak (changes requested):

packages-ts/gauntlet-terra/src/commands/internal/terra.ts
RodrigoAD: Yes. The list is slowly growing so makes sense

packages-ts/gauntlet-terra/src/commands/internal/terra.ts
RodrigoAD: Yeah right, is not used. Let me expose it. I'll leave it as `final`

packages-ts/gauntlet-terra/src/commands/internal/terra.ts
RodrigoAD: For now we have no expectation of using this, I'll wait for implementing the command

packages-ts/gauntlet-terra-contracts/src/commands/contracts/migration/index.ts
```diff
@@ -0,0 +1,84 @@
...
+  contract: string
+  newCodeId: number
+  migrationMsg: MigrationMessage
+}
+
+export type MigrationMessageMaker<MigrationMessage> = (flags, args) => MigrationMessage
+
+export const makeMigrationCommand = <MigrationMessage>(
+  contract: CONTRACT_LIST,
+  makeMigrationMessage: MigrationMessageMaker<MigrationMessage>,
```
RodrigoAD, line 17: I went a bit ahead, we'll need this command to be particular for each contract as every migration message will be different.
The only thing each contract command has to do is passing `makeMigratioMessage` that will construct the message based on the user input. 
I made an example on how it should look like with the OCR contract: https://github.com/smartcontractkit/chainlink-terra/pull/219/files#diff-f5e9f3694381cb803e8c4fa7798f04c6e87e3adaa92c9e9e6f184a19952a82c3R7

RodrigoAD:
This should be ready @krebernisak

packages-ts/gauntlet-terra-cw-plus/src/commands/multisig.ts
```diff
@@ -1,16 +1,19 @@
...
 import { isDeepEqual } from '../lib/utils'
 import { fetchProposalState, makeInspectionMessage } from './inspect'
 import { Vote, Cw3WasmMsg, Action, State, Cw3BankMsg } from '../lib/types'
 
-type ProposalAction = (
-  signer: AccAddress,
-  proposalId: number,
-  message: MsgExecuteContract | MsgSend,
-) => Promise<MsgExecuteContract>
+type TerraMessage = MsgExecuteContract | MsgSend | MsgUpdateContractAdmin | MsgMigrateContract
```
krebernisak, line 15: Isn't a type like this already available in `'@terra-money/terra.js'`?

packages-ts/gauntlet-terra-cw-plus/src/commands/multisig.ts
```diff
@@ -154,7 +181,7 @@ export const wrapCommand = (command) => {
       // TODO: Gauntlet core should initialize commands using `buildCommand` instead of new Command
       await this.buildCommand(this.flags, this.args)
 
-      let proposalId = Number(this.flags.proposal || this.flags.multisigProposal) // alias requested by eng ops
+      let proposalId = Number(this.flags.proposal || this.flags.multisigProposal)
```
krebernisak, line 184: These comments had a purpose:

- Point to alias as an external feature request
- Point that aliases are not explicitly supported by Gauntlet and are added ad-hoc

I'd suggest tracking this somewhere and figuring out how aliases fit the bigger picture and how are they handled by the framework. I'm personally against any kind of aliases at this point as they bring more confusion than clarity.

packages-ts/gauntlet-terra/src/commands/internal/terra.ts
krebernisak: Could you please add a comment that this is an unused piece of code that needs to be moved to its own command?

packages-ts/gauntlet-terra/src/commands/internal/terra.ts
krebernisak: This was not moved to the `MigrateContract`, why not?

packages-ts/gauntlet-terra-contracts/src/commands/tooling/updateContractAdmin.ts
```diff
@@ -0,0 +1,95 @@
...
+import { AccAddress, MsgUpdateContractAdmin } from '@terra-money/terra.js'
+import { CATEGORIES } from '../../lib/constants'
+
+type CommandInput = {
+  newAdmin: string
+  contract: string
+  force?: boolean
+}
+
+export default class UpdateContractAdmin extends TerraCommand {
```
krebernisak, line 13: This command has nothing to do with our contracts. We can update contract admin for any contract, not just Chainlink contracts.

This should be moved to `gauntlet-terra` base package, so either try to do that now or add an inline comment noting the next steps oh how to restructure this.

packages-ts/gauntlet-terra-contracts/src/commands/contracts/migration/index.ts
```diff
@@ -0,0 +1,84 @@
...
+) => {
+  return class MigrateContract extends TerraCommand {
```
krebernisak, line 19: This command has nothing to do with our contracts. We should build functionality to migrate any contract, not just Chainlink contracts in this package.

This should be moved to `gauntlet-terra` base package, so either try to do that now or add an inline comment noting the next steps oh how to restructure this.

krebernisak, line 19: There is a lot of smart programming going on here that makes things harder to follow. I'd suggest defining a class as a top-level type, and then you could have a static constructor on it or a separate factory class/fn.

Not really fond of defining a (large) class in a closure like this.

packages-ts/gauntlet-terra/src/commands/internal/terra.ts
krebernisak: Aha, this is not used anymore?

packages-ts/gauntlet-terra-contracts/src/commands/contracts/migration/index.ts
RodrigoAD: Yep I see your point, this makes the Class inaccessible from outside. Let me fix it

packages-ts/gauntlet-terra-cw-plus/src/commands/multisig.ts
RodrigoAD: Yeah you are right. Updating it to support every Message type

packages-ts/gauntlet-terra-contracts/src/commands/contracts/ocr2/migrate.ts
```diff
@@ -0,0 +1,21 @@
...
+// export default makeMigrationCommand(CONTRACT_LIST.OCR_2, makeMigrateMessage)
+
+export default class OCRMigrateCommand extends MigrateContract<OCR2MigrateMessage> {
+  static description = MigrateContract.makeDescription(CONTRACT_LIST.OCR_2)
+  static examples = MigrateContract.makeExamples(CONTRACT_LIST.OCR_2)
+
+  static id = MigrateContract.makeId(CONTRACT_LIST.OCR_2)
+  static category = MigrateContract.makeCategory(CONTRACT_LIST.OCR_2)
+
+  makeMigrationMessage = makeMigrationMessage
```
krebernisak, line 20: Let's declare the abstract fn here instead of assigning it to it like this.

packages-ts/gauntlet-terra-contracts/src/commands/contracts/ocr2/migrate.ts
```diff
@@ -0,0 +1,21 @@
+import { CONTRACT_LIST } from '../../../lib/contracts'
+import { MigrateContract, MigrationMessageMaker } from '../migration'
+
+// TODO: Pending of migrate function on the contract. Serve this as an example of implementation
+type OCR2MigrateMessage = any
```
krebernisak, line 5: Let's test this and update this comment.

Also, the migration should be versioned somehow. At least in the message. For example, migration from `v0.1.5` to `v1.0.0` requires an Empty message, but the next migration to `v1.1.0` could require a specific non-empty message.

Not sure if we should support more than just the latest migration though. But at least let's be explicit about the version of the message, even if we declare just one (latest).

```suggestion
// Notice: Msg used to migrate OCR2 from v0.1.5 to v1.0.0
type MsgMigrate_1_0_0 = any
```

packages-ts/gauntlet-terra-contracts/src/commands/contracts/migration/index.ts
```diff
@@ -0,0 +1,81 @@
...
+  migrationMsg: MigrationMessage
+}
+
+export type MigrationMessageMaker<MigrationMessage> = (flags, args) => MigrationMessage
+
+export abstract class MigrateContract<MigrationMessage> extends TerraCommand {
+  static makeDescription = (contract) =>
+    `(NOT TESTED!) Upgrades a contract ${contract} instance to use a new Code ID. This new Code ID must expose a "migrate" function to migrate the contract state`
+  static makeExamples = (contract) => [
+    `yarn gauntlet ${contract}:migrate_contract --network=bombay-testnet --newCodeId=<CODE_ID_NUMBER> <CONTRACT_ADDRESS>`,
```
krebernisak, line 19:
```suggestion
    `yarn gauntlet ${contract}:migrate_contract --network=<NETWORK> --newCodeId=<CODE_ID_NUMBER> <CONTRACT_ADDRESS>`,
```

packages-ts/gauntlet-terra-contracts/src/commands/contracts/migration/index.ts
```diff
@@ -0,0 +1,81 @@
+import { Result } from '@chainlink/gauntlet-core'
+import { prompt } from '@chainlink/gauntlet-core/dist/utils'
+import { TerraCommand, TransactionResponse } from '@chainlink/gauntlet-terra'
+import { AccAddress, MsgMigrateContract } from '@terra-money/terra.js'
+import { CONTRACT_LIST } from '../../../lib/contracts'
+
+type CommandInput<MigrationMessage> = {
```
krebernisak, line 7: From `@terra-money/terra.js` we get message types abbreviated as `Msg`, so let's use the same in our own types to be consistent.

```suggestion
type CommandInput<MsgMigrate> = {
```

packages-ts/gauntlet-terra-contracts/src/commands/contracts/migration/index.ts
```diff
@@ -0,0 +1,81 @@
...
+  contract: string
+  newCodeId: number
+  migrationMsg: MigrationMessage
```
krebernisak, line 10:
```suggestion
  msgMigrate: MsgMigrate
```

packages-ts/gauntlet-terra-contracts/src/commands/contracts/migration/index.ts
```diff
@@ -0,0 +1,81 @@
...
+}
+
+export type MigrationMessageMaker<MigrationMessage> = (flags, args) => MigrationMessage
+
+export abstract class MigrateContract<MigrationMessage> extends TerraCommand {
+  static makeDescription = (contract) =>
+    `(NOT TESTED!) Upgrades a contract ${contract} instance to use a new Code ID. This new Code ID must expose a "migrate" function to migrate the contract state`
```
krebernisak, line 17: Let's test it:

```suggestion
    `Upgrades a contract ${contract} instance to use a new Code ID. This new Code ID must expose a "migrate" function to migrate the contract state`
```

packages-ts/gauntlet-terra-contracts/src/commands/contracts/ocr2/migrate.ts
```diff
@@ -0,0 +1,21 @@
...
+
+const makeMigrationMessage: MigrationMessageMaker<OCR2MigrateMessage> = (flags, args) => {
+  return {}
+}
+
+// export default makeMigrationCommand(CONTRACT_LIST.OCR_2, makeMigrateMessage)
+
+export default class OCRMigrateCommand extends MigrateContract<OCR2MigrateMessage> {
```
krebernisak, line 13: First, notice the `OCR` vs `OCR2` prefix.
Second, I don't think we should namespace like this as we can always `import {A as B}` It's already clear this command is in OCR2 context as it's located under `/contracts/ocr2/`.

packages-ts/gauntlet-terra-contracts/src/commands/tooling/updateContractAdmin.ts
```diff
@@ -0,0 +1,95 @@
...
+  input: CommandInput
+
+  static description = 'Updates contract admin. Admin role can migrate the contract to a new Code ID'
+  static examples = [
+    `yarn gauntlet tooling:update_contract_admin --network=bombay-testnet --to=<NEW_ADMIN> <CONTRACT_ADDRESS>`,
```
krebernisak, line 18:
```suggestion
    `yarn gauntlet tooling:update_contract_admin --network=<NETWORK> --to=<NEW_ADMIN> <CONTRACT_ADDRESS>`,
```

krebernisak (changes requested):

packages-ts/gauntlet-terra-contracts/src/commands/contracts/migration/index.ts
```diff
@@ -0,0 +1,80 @@
+import { Result } from '@chainlink/gauntlet-core'
+import { prompt } from '@chainlink/gauntlet-core/dist/utils'
+import { TerraCommand, TransactionResponse } from '@chainlink/gauntlet-terra'
+import { AccAddress, MsgMigrateContract } from '@terra-money/terra.js'
+
+type CommandInput<MsgMigrate> = {
+  contract: string
+  newCodeId: number
+  migrationMsg: MsgMigrate
```
krebernisak, line 9: nit: terra.js has a consistent `Msg` prefix for all messages, so let's also try to be consistent here and start with the `msg/Msg` prefix for all vars, members, and types.

```suggestion
  msgMigrate: MsgMigrate
```

packages-ts/gauntlet-terra-contracts/src/commands/contracts/migration/index.ts
```diff
@@ -0,0 +1,80 @@
...
+}
+
+export type MigrationMessageMaker<MsgMigrate> = (flags, args) => MsgMigrate
+
+export abstract class MigrateContract<MsgMigrate> extends TerraCommand {
+  static makeDescription = (contract) =>
+    `(NOT TESTED!) Upgrades a contract ${contract} instance to use a new Code ID. This new Code ID must expose a "migrate" function to migrate the contract state`
+  static makeExamples = (contract) => [
+    `yarn gauntlet ${contract}:migrate_contract --network=<NETWORK> --newCodeId=<CODE_ID_NUMBER> <CONTRACT_ADDRESS>`,
+    `yarn gauntlet ${contract}:migrate_contract --network=mainnet --newCodeId=2012 terra167ccv2h0z7k0p8j6qpuzwsgu5au5qvfwgmkjsl`,
```
krebernisak, line 19: Do we need these specific non-general examples? I assume they are helpful to communicate types for different parameters, but what we should do instead is explicitly add typings to the framework, so help for command would list all the available flags/args and their descriptive types.

Something to consider going forward.

packages-ts/gauntlet-terra-contracts/src/commands/contracts/migration/index.ts
```diff
@@ -0,0 +1,80 @@
...
+
+  makeInput = (flags, args): CommandInput<MsgMigrate> => {
+    return {
+      newCodeId: Number(flags.newCodeId),
+      contract: args[0],
+      migrationMsg: this.makeMigrationMessage(flags, args),
+    }
+  }
+
+  validateInput = (input: CommandInput<MsgMigrate>): boolean => {
```
krebernisak, line 50: Consider changing this to:

```suggestion
  requireValidInput = (input: CommandInput<MsgMigrate>) => {
```

as this is more explicit to how it's currently used.

krebernisak (approved):
Looks ready to be merged after it's tested against the contracts migrate PR branch:

- https://github.com/smartcontractkit/chainlink-terra/pull/232

PR #222, opened 2022-03-16
Bugfix: wrong answers and config digests

topliceanu:


pkg/monitoring/source_envelope.go
```diff
@@ -275,35 +276,81 @@ func (e *envelopeSource) fetchLinkBalance(ctx context.Context) (*big.Int, error)
...
+	eventType string,
+	contractAddressBech32 string,
+	res *cosmosTx.GetTxsEventResponse,
+	extractors map[string]func(string) error,
+) error {
 	if len(res.TxResponses) == 0 ||
 		len(res.TxResponses[0].Logs) == 0 ||
 		len(res.TxResponses[0].Logs[0].Events) == 0 {
-		return fmt.Errorf("no events found of event type '%s'", eventType)
+		return fmt.Errorf("%d events found in response", len(res.TxResponses[0].Logs[0].Events))
```
jmank88, line 288: This looks like it will nil panic for the first two len checks.

jmank88 (approved):

PR #226, opened 2022-03-17
Fuzz and unit test median from report

connorwstein:


pkg/terra/report_test.go
```diff
@@ -78,4 +79,73 @@ func TestMedianFromReport(t *testing.T) {
...
+		obs            []*big.Int
+		expectedMedian *big.Int
+	}{
+		{
+			name:           "2 positive one zero",
+			obs:            []*big.Int{big.NewInt(0), big.NewInt(10), big.NewInt(20)},
+			expectedMedian: big.NewInt(10),
+		},
+		{
+			name:           "2 positive one zero",
```
krebernisak, line 98: This test is a duplicate of the previous one.

krebernisak (dismissed):
Looks great!

While it's fresh, it would be awesome if you can replicate the same kind of test on Solana 🙏

pkg/terra/report_fuzz_test.go
```diff
@@ -0,0 +1,35 @@
...
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/smartcontractkit/libocr/offchainreporting2/reportingplugin/median"
+)
+
+// go test -tags=go1.18 -fuzz ./...
+func FuzzMedianFromReport(f *testing.F) {
```
jmank88, line 17: Should the func name have `ReportCode` too?
```suggestion
func FuzzReportCodecMedianFromReport(f *testing.F) {
```

jmank88 (approved):

krebernisak (approved):

PR #230, opened 2022-03-21
update juelsPerFeeCoin spec

aalu1418:


krebernisak (approved):
Great resolution @aalu1418!

We should really make these examples less static, and actually use them as templates in tests so we make sure they are always up to date, and to surface any issues early.

aalu1418:
> Great resolution @aalu1418!
> 
> We should really make these examples less static, and actually use them as templates in tests so we make sure they are always up to date, and to surface any issues early.

added an issue to track! https://github.com/smartcontractkit/chainlink-terra/issues/231

PR #232, opened 2022-03-22
Migrate

archseer:


contracts/ocr2/src/contract.rs
```diff
@@ -259,6 +258,32 @@ pub fn reply(_deps: DepsMut, _env: Env, msg: Reply) -> Result<Response, Contract
...
 }
 
+// Migrate contract if version is lower than current version
+#[cfg_attr(not(feature = "library"), entry_point)]
+pub fn migrate(deps: DepsMut, _env: Env, _msg: Empty) -> Result<Response, ContractError> {
+    let ver = cw2::get_contract_version(deps.storage)?;
+
+    // ensure we are migrating from an allowed contract
+    if ver.contract != CONTRACT_NAME {
+        return Err(cosmwasm_std::StdError::generic_err("Can only upgrade from same type").into());
```
krebernisak, line 268: Why don't we explicitly define a `MigrationTypeMismatch` error in `error.rs`.

krebernisak (approved):

contracts/ocr2/src/contract.rs
archseer: Doesn't really matter since we're not catching the error anywhere, it's just used in `tx migrate`. Can change it though

krebernisak: I have a slight preference to change it, as it makes the error more explicit and discoverable. At the same time code is more readable (consider tests) and typed.

Additionally, I guess at one point we could add a base crate and share some of these base types and functionality like we currently do with the `Owned` crate (sharing the `Unauthorized` error in this context).

It is a nitpick ATM though, so feel free to go either direction for now.

krebernisak (dismissed):

contracts/ocr2/src/contract.rs
```diff
@@ -259,6 +258,34 @@ pub fn reply(_deps: DepsMut, _env: Env, msg: Reply) -> Result<Response, Contract
...
+    let stored = cw2::get_contract_version(deps.storage)?;
+
+    // ensure we are migrating from an allowed contract
+    if stored.contract != CONTRACT_NAME {
+        return Err(ContractError::CannotMigrate {
+            previous_contract: stored.contract,
+        });
+    }
+
+    // TODO: could use semver to only allow upgrades:
```
krebernisak, line 273: Why is this TODO useful here?

The code of the migrate fn is set by the user executing the migration, so as they can add/remove any checks we can't really enforce this, right?

contracts/ocr2/src/contract.rs
```diff
@@ -1472,8 +1499,8 @@ fn calculate_reimbursement(
     signature_count: usize,
 ) -> Uint128 {
     let signature_count = signature_count as u64;
-    let gas_per_signature = config.gas_per_signature.unwrap_or(17_000);
-    let gas_base = config.gas_base.unwrap_or(84_000);
+    let gas_per_signature = config.gas_per_signature.unwrap_or(4_096);
+    let gas_base = config.gas_base.unwrap_or(146_000);
     let gas_adjustment = Decimal::percent(u64::from(config.gas_adjustment.unwrap_or(140)));
```
krebernisak, line 1504: We synced offline on the topic of "On chain 1.4x multiplier produces matching values to 1.5x offchain"...

Let's add a descriptive in-line comment here explaining why are the values different, and where is the discrepancy.

krebernisak (dismissed):

krebernisak (approved):

Issue #78, opened 2022-02-03
Update local testing environment for correct chain + node configs


aalu1418:
The local testing environment passes the FCD client for gas calculations in the node config rather than the chain config. 

The CL node expects that it is passed via the chain config. Additional changes are likely needed in:
* integrations-framework
* chainlink-relay/ops

krebernisak:
This should be fixed now.

PR #233, opened 2022-03-22
e2e tests: Observation source for every contract

gheorghestrimtu:


tests/e2e/common/common.go
```diff
@@ -168,50 +168,56 @@ func DefaultOffChainConfigParamsFromNodes(nodes []client.Chainlink) (contracts.O
...
+			}
+			_, err = n.CreateTerraChain(&client.TerraChainAttributes{ChainID: "localterra"})
+			if err != nil {
+				return nil, err
+			}
+			if _, err = n.CreateTerraNode(&client.TerraNodeAttributes{
+				Name:          "terra",
+				TerraChainID:  relayConfig["chainID"],
+				TendermintURL: relayConfig["tendermintURL"],
+				FCDURL:        relayConfig["fcdURL"],
```
krebernisak, line 213: `FCDURL` is not part of the node's config (represented by `TerraNodeAttributes`) anymore (for a while now), but actually a part of chains config (represented by `TerraChainAttributes`).

krebernisak (changes requested):

tests/e2e/common/common.go
krebernisak: Here is the relevant issue @gheorghestrimtu:

-  https://github.com/smartcontractkit/chainlink-terra/issues/78

krebernisak: It would be great if you can move this along as part of this effort 🙌

gheorghestrimtu: Of course, no problem

tests/e2e/smoke/common/common.go
```diff
@@ -220,13 +228,19 @@ func (m *OCRv2State) CreateJobs() {
 	m.SetAllAdapterResponsesToTheSameValue(5)
 	err := m.MockServer.SetValuePath("/juels", 1)
 	Expect(err).ShouldNot(HaveOccurred())
-	bi, err := common.CreateBridges(m.Nodes, m.MockServer)
+	ocr2Contracts := []string{}
+	for _, contract := range m.Contracts {
+		ocr2Contracts = append(ocr2Contracts, contract.OCR2.Address())
+	}
+	err = common.CreateTerraChainAndNode(m.Nodes)
+	Expect(err).ShouldNot(HaveOccurred())
```
krebernisak, line 236: nit: switch the order here
```suggestion
        err = common.CreateTerraChainAndNode(m.Nodes)
	Expect(err).ShouldNot(HaveOccurred())

	ocr2Contracts := []string{}
	for _, contract := range m.Contracts {
		ocr2Contracts = append(ocr2Contracts, contract.OCR2.Address())
	}
```

krebernisak (approved):
LGTM!

It'd be great if you could check the consistency with Solana e2e test implementation, and do the same changes there 🙌

krebernisak (approved):

PR #236, opened 2022-03-25
pkg/terra: support replacing messages

jmank88:
https://app.shortcut.com/chainlinklabs/story/32618/tx-strategies-to-handle-queue-pruning
Supports: https://github.com/smartcontractkit/chainlink/pull/6301

### DIFF
--- a/pkg/terra/db/db.go
+++ b/pkg/terra/db/db.go
@@ -68,7 +68,10 @@ var (
 	// Confirmed means we're able to retrieve the txhash of the tx which broadcasted the msg.
 	// Valid next states: none, terminal state
 	Confirmed State = "confirmed"
-	// Errored means the msg reverted in simulation OR the tx containing the message timed out waiting to be confirmed
+	// Errored means the msg:
+	//  - reverted in simulation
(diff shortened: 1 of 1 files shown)


pkg/terra/db/db.go
```diff
@@ -60,14 +60,17 @@ type State string
...
 	Unstarted State = "unstarted"
 	// Broadcasted means included in the mempool of a node.
 	// Valid next states: Confirmed (found onchain), Errored (tx expired waiting for confirmation)
 	Broadcasted State = "broadcasted"
 	// Confirmed means we're able to retrieve the txhash of the tx which broadcasted the msg.
 	// Valid next states: none, terminal state
 	Confirmed State = "confirmed"
+	// Replaced means the caller sent another message which superseded this one before it was broadcast.
+	// Valid next states, none, terminal state. However, a mid-broadcast replacement attempt could be overwritten.
+	Replaced State = "replaced"
```
jmank88, line 73: I'm thinking this would be simpler and more flexible going forward if generalized, e.g. `cancelled`, with the particular case of replacement being an implementation detail.
```suggestion
	Cancelled State = "cancelled"
```

PR #237, opened 2022-03-25
pkg/terra/client: lower malformed json log level from error to warn

jmank88:
https://app.shortcut.com/chainlinklabs/story/34069/graceful-handling-of-errors-invalid-json-from-bison-trails


jmank88:
I hacked up a version to restore the id when it is missing, but it's pretty messy would introduce extra latency on every request (no just invalid).

krebernisak (approved):

PR #240, opened 2022-03-28
Add responses command for ocr2

dimriou:


packages-ts/gauntlet-terra-contracts/src/commands/contracts/ocr2/inspection/responses.ts
```diff
@@ -0,0 +1,200 @@
...
+import { CATEGORIES } from '../../../../lib/constants'
+import { LCDClient } from '@terra-money/terra.js'
+import { getLatestOCRNewTransmissionEvent, parseObservers, RoundData } from '../../../../lib/inspection'
+import { inspection, logger } from '@chainlink/gauntlet-core/dist/utils'
+
+type CommandInput = {
+  transmitters: string[]
+  description: string
+  operators: string[]
+  aggregatorOracles: any[]
```
RodrigoAD, line 13: Be more explicit on the Input. What does the user need to provide exactly? Imagine the commandInput as a JSON the user has to provide for the command to perfectly work.

I also see some overlapping between transmitters, operators and aggregatorOracles. We could have a type that is `Oracle` that has all the oracle needed info.
If the user wants us to make the command input from RDD, that's fine, but the rest of the command (like the `inspect` function) should not be aware of RDD anymore.

packages-ts/gauntlet-terra-contracts/src/commands/contracts/ocr2/inspection/responses.ts
```diff
@@ -0,0 +1,200 @@
...
+      roundId: latestRoundData.round_id,
+      answer: latestRoundData.answer,
+      observationsTimestamp: latestRoundData.observations_timestamp,
+      transmissionTimestamp: latestRoundData.transmission_timestamp,
+    },
+    event,
+  }
+}
+
+const inspect = (expected: ContractExpectedInfo, onchainData: ContractExpectedInfo): boolean => {
```
RodrigoAD, line 88: Try to refactor this function using a better structured Input, as I suggested in the previous comment. 
Remember that here there's no concept of RDD anymore, just the user Input. RDD is just a source that we use to generate an Input.

packages-ts/gauntlet-terra-contracts/src/commands/contracts/ocr2/inspection/responses.ts
```diff
@@ -0,0 +1,195 @@
...
+import { CONTRACT_LIST } from '../../../../lib/contracts'
+import { CATEGORIES } from '../../../../lib/constants'
+import { LCDClient } from '@terra-money/terra.js'
+import { getLatestOCRNewTransmissionEvent, parseObserversByLength, RoundData } from '../../../../lib/inspection'
+import { inspection, logger } from '@chainlink/gauntlet-core/dist/utils'
+
+type CommandInput = {
+  transmitters: string[]
+  description: string
+  observersInfo: { [address: string]: ObserverInfo }
```
RodrigoAD, line 12: We are splitting the oracle information between `transmitters` and `observers`, but at this point, we don't know who are transmitters and who are the observers. Let's just have one `oracles` that contains all the info necessary:
```js
type CommandInput = {
  transmitters: string[]
  description: string
  oracles: Oracle[]
}

type Oracle = {
  transmitter: string
  name: string
  apis: string
}
```

packages-ts/gauntlet-terra-contracts/src/lib/inspection.ts
```diff
@@ -12,3 +19,23 @@ export const getLatestOCRConfigEvent = async (provider: LCDClient, contract: Acc
...
+      ],
+      'pagination.limit': '1',
+      order_by: 'ORDER_BY_DESC',
+    })
+  ).txs
+
+  return transmissionTx.length > 0 ? transmissionTx[0]?.logs?.[0].eventsByType['wasm-new_transmission'] : null
+}
+
+export const parseObserversByLength = (observers: string, observersNumber: number): number[] =>
```
RodrigoAD, line 40: Bring this function closer where it is used. If in the future is useful in somewhere else, we could bring it to a common place then

packages-ts/gauntlet-terra-contracts/src/lib/inspection.ts
```diff
@@ -12,3 +19,23 @@ export const getLatestOCRConfigEvent = async (provider: LCDClient, contract: Acc
 
   return setConfigTx?.logs?.[0].eventsByType['wasm-set_config']
 }
+
+export const getLatestOCRNewTransmissionEvent = async (provider: LCDClient, contract: AccAddress) => {
+  let transmissionTx = (
+    await provider.tx.search({
```
RodrigoAD, line 25: I think this is complex enough to make a helper, like `getLatestContractEvent = (provider, event: string, contract: string)` ?

packages-ts/gauntlet-terra-contracts/src/lib/inspection.ts
```diff
@@ -12,3 +19,23 @@ export const getLatestOCRConfigEvent = async (provider: LCDClient, contract: Acc
...
+          key: 'wasm-new_transmission.contract_address',
+          value: contract,
+        },
+      ],
+      'pagination.limit': '1',
+      order_by: 'ORDER_BY_DESC',
+    })
+  ).txs
+
+  return transmissionTx.length > 0 ? transmissionTx[0]?.logs?.[0].eventsByType['wasm-new_transmission'] : null
```
RodrigoAD, line 37: Will return undefined of the `length` is zero
```suggestion
  return transmissionTx[0]?.logs?.[0].eventsByType['wasm-new_transmission']
```

packages-ts/gauntlet-terra-contracts/src/commands/contracts/ocr2/inspection/responses.ts
```diff
@@ -0,0 +1,195 @@
...
+  if (!event) {
+    logger.error(
+      `Could not find NewTransmission event of aggregator ${onchainData.aggregatorAddress} for the latest round).`,
+    )
+    return false
+  }
+  logger.debug(event)
+  logger.debug(onchainData.transmitters)
+
+  const roundTime = new Date(Number(event.observations_timestamp) * 1000)
```
RodrigoAD, line 115: We can use the `dateFromUnix` helper here

packages-ts/gauntlet-terra-contracts/src/commands/contracts/ocr2/inspection/responses.ts
```diff
@@ -0,0 +1,195 @@
...
+  if (Object.keys(unresponsiveNodes).length > 0) {
+    logger.warn(`Found ${Object.keys(unresponsiveNodes).length} unresponsive nodes:`)
+    logger.log(unresponsiveNodes)
+  } else {
+    logger.success(`All ${onchainData.transmitters.length} nodes responded.`)
+  }
+
+  logger.info(`Aggregated answer in round ${event.aggregator_round_id}: ${event.answer}`)
+
+  return false
```
RodrigoAD, line 163: Inspect should return is the inspection was successful. If we reached this point, the transmission was ok, so we just need to run the other inspections
```suggestion
  return inspection.inspect(inspections)
```

packages-ts/gauntlet-terra-contracts/src/commands/contracts/ocr2/inspection/responses.ts
```diff
@@ -0,0 +1,195 @@
...
+
+  const roundTime = new Date(Number(event.observations_timestamp) * 1000)
+  logger.info(`Found NewTransmission event for round ${event.aggregator_round_id} (${roundTime})`)
+
+  let responsiveNodes: { [address: string]: ObserverInfo } = {}
+  let unresponsiveNodes: { [address: string]: ObserverInfo } = {}
+
+  const parsedObserversIndexes = parseObserversByLength(event.observers, event.observations.length)
+
+  parsedObserversIndexes.forEach((val, i) => {
```
RodrigoAD, line 123: This and [next](inspection.inspect(inspections)) are hard to understand. 
- Use a simpler data structure. Maybe only have one `responsiveNodes: Oracle[]`
- `unresponsiveNodes` could just be `unresponsiveNodes = input.oracles.filter((o) => !responsiveNodes.includes(o))`
- Try to declare the variables along to their values, that way we avoid mutability and increase readability 
  - `const responsiveNodes = parsedObserversIndexes.map(() => ...)`

dimriou: I initially had implemented it as you describe above, but I noticed there was too much searching in lists(filter,map, find). Although the size is too small to even be significant for performance,  I tried to have a general structure from which you can directly find elements' presence. This is also why for the CommandInput type I opted for a dictionary rather than a single Observer/Oracle object.
I agree though that readability took a hit with that.

packages-ts/gauntlet-terra-contracts/src/commands/contracts/ocr2/inspection/responses.ts
```diff
@@ -0,0 +1,188 @@
...
+const makeOnchainData = (provider: LCDClient) => async (
+  instructionsData: any[],
+  input: CommandInput,
+  aggregator: string,
+): Promise<ContractExpectedInfo> => {
+  const latestRoundData = instructionsData[0]
+  const transmitters = instructionsData[1]
+  const description = instructionsData[2]
+
+  let event
```
RodrigoAD, line 73: Be careful here, event will have an `any` type here, meaning that you can include anything you want there. A way to fix this could be:
```js
      ...
      transmissionTimestamp: latestRoundData.transmission_timestamp,
    },
    ...(onchainEvent && {
      event: {
        answer: Number(onchainEvent.answer[0]),
        transmitter: onchainEvent.transmitter[0],
        observations: onchainEvent.observations.map((o) => Number(o)),
        observers: onchainEvent.observers[0],
        observations_timestamp: Number(onchainEvent.observations_timestamp[0]),
        aggregator_round_id: Number(onchainEvent.aggregator_round_id[0]),
      },
    }),
```
And marking `event` as optional in `ContractExpectedInfo`

packages-ts/gauntlet-terra-contracts/src/commands/contracts/ocr2/inspection/responses.ts
```diff
@@ -0,0 +1,188 @@
...
+    )
+    return false
+  }
+  logger.debug(event)
+  logger.debug(onchainData.transmitters)
+
+  const roundTime = dateFromUnix(event.observations_timestamp)
+  logger.info(`Found NewTransmission event for round ${event.aggregator_round_id} (${roundTime})`)
+
+  const parseObserversByLength = (observers: string, observersNumber: number): number[] =>
```
RodrigoAD, line 119: Better to bring this outside of the `inspect` fn

RodrigoAD (commented):
Just 2 final comments, we can merge after

RodrigoAD (approved):
LGTM! Thanks @dimriou . Feel free to merge when CI passes

RodrigoAD (approved):

PR #241, opened 2022-03-29
Add Explicit Names For CI

tateexon:


.github/workflows/chaos.yml
```diff
@@ -11,8 +11,8 @@ on:
         type: string
 
 jobs:
-  build_custom_cl_image:
-    name: Build custom CL image
+  E2E_Chaos_Build_Custom_Chainlink_Image:
```
krebernisak, line 14: Do these actually show anywhere now that we have explicit names for all?

I'd prefer we leave them lowercase, to follow the rest of `yml` workflow style, but we should make sure to add an explicit name for each.

.github/workflows/gauntlet.yml
```diff
@@ -26,9 +24,10 @@ jobs:
           nix develop -c yarn install --frozen-lockfile
         env:
           NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
```
krebernisak, line 26: Please also remove the `NPM_TOKEN` setup from workflows, as `gautlet-core` should now be public so this shouldn't be necessary anymore.

.github/workflows/golangci-lint.yml
```diff
@@ -3,14 +3,16 @@ name: golangci-lint
 on:
   push:
     branches:
-      - auto
-      - try
-      - rollup
+      - develop
+      - main
+    paths:
+      - '**.go'
```
krebernisak, line 9: Will this only run if `'**.go'` files are touched?

If that's the case, then we can't enable this as a global check for develop branch. Consider removing this filter, for now, so we enable this check globally.

krebernisak (approved):

krebernisak (changes requested):

.github/workflows/gauntlet.yml
```diff
@@ -1,13 +1,11 @@
...
 on:
   push:
-    branches:
-      - develop
-  pull_request:
 
 jobs:
-  build:
+  gauntlet_build:
+    name: Guantlet Build
```
aalu1418, line 8:
```suggestion
    name: Gauntlet Build
```

aalu1418 (changes requested):

.github/workflows/gauntlet.yml
tateexon: 🤦

aalu1418 (approved):

.github/workflows/chaos.yml
```diff
@@ -1,4 +1,4 @@
-name: E2E Chaos Tests
+name: e2e_chaos_tests
```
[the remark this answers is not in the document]
krebernisak: Why the discrepancy from Solana? What was the problem when using nix there?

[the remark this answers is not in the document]
tateexon: Nix on Solana had a 1 hour cache issue for the first call using it for every run. We don't have that for Terra. When we solve that cache issue for nix on Solana have it use nix as well.

aalu1418 (approved):

.golangci.yml
```diff
@@ -37,7 +37,7 @@ linters-settings:
       - name: error-return
       - name: error-strings
       - name: error-naming
-      - name: exported
+      # - name: exported
```
krebernisak, line 40: We should document all Go exported (public) types

krebernisak (approved):

PR #243, opened 2022-03-30
Add namespace to proxy answer and link available metrics

pedrovsg:


topliceanu (approved):
Note that this is going to break existing charts at least in here https://chainlinklabs.grafana.net/d/OGyagufnz/onchain-monitoring-tom-som?orgId=1&from=now-2d&to=now&var-job=atlas_prod_tom_terra-mainnet&var-logs=atlas-prod-atlas-solana-onchain-monitor-mainnet&var-feed_name=ATOM%20%2F%20USD

topliceanu:
I was thinking, maybe for consistency we could use the namespace feature and add a namespace for all these metrics. Wdyt?

PR #258, opened 2022-04-13
packages-ts/terra-sdk: add blockHeight, txIndex, txHash to Round

jmank88:
https://app.shortcut.com/chainlinklabs/story/36233/terra-sdk-solana-sdk-add-chain-timestamp-to-round

On-chain timestamp was requested, but that doesn't seem to be available on the subscription response. We do have block height though, as well as tx index and hash.

### DIFF
--- a/packages-ts/terra-sdk/src/index.ts
+++ b/packages-ts/terra-sdk/src/index.ts
@@ -8,6 +8,10 @@ export interface Round {
   epoch: number
   aggregatorRoundId: number
   observationsTS: Date
+
+  blockHeight: bigint
+  txIndex: bigint
+  txHash: string
 }
 
 export class OCR2Feed {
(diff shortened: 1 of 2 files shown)


packages-ts/terra-sdk/src/index.ts
```diff
@@ -8,6 +8,10 @@ export interface Round {
   epoch: number
   aggregatorRoundId: number
   observationsTS: Date
+
+  blockHeight: bigint
+  txIndex: bigint
+  txHash: string
```
jmank88, line 14: @krebernisak We could also add the timestamp, but IIUC that would require extra calls. I think I would like to keep this simple and efficient, but we could provide a helper for fetching additional info if we need to.

archseer (approved):

PR #259, opened 2022-04-13
Populate config synchronously on start

connorwstein:


pkg/terra/contract_cache.go
```diff
@@ -48,6 +48,13 @@ func NewContractCache(cfg Config, reader *OCR2Reader, lggr Logger) *ContractCach
 }
 
 func (cc *ContractCache) Start() error {
+	// We synchronously update the config on start so that
+	// when OCR starts there is config available (if possible).
+	// Avoids confusing "contract has not been configured" OCR errors.
+	ctx, _ := utils.ContextFromChan(cc.stop)
```
jmank88, line 54: Does this leak a goroutine if you don't call cancel?

connorwstein: yep looks like, will fix elsewhere too

jmank88 (approved):

PR #264, opened 2022-04-19
Update references to AddressBook

akhilchainani:


packages-ts/gauntlet-terra-contracts/src/lib/middlewares.ts
```diff
@@ -1,5 +1,6 @@
-import { TerraCommand, AddressBook, logger } from '@chainlink/gauntlet-terra'
+import { TerraCommand, logger } from '@chainlink/gauntlet-terra'
 import { Middleware, Next } from '@chainlink/gauntlet-core'
+import { AddressBook } from '@chainlink/gauntlet-core/dist/commands/addressBook'
```
RodrigoAD: Let's make the change on `gauntlet-core` first, make the release, and then update `gauntlet-terra`

packages-ts/gauntlet-terra/src/commands/logger.ts
```diff
@@ -1,5 +1,5 @@
 import { logger } from '@chainlink/gauntlet-core/dist/utils'
-import { AddressBook } from './addressBook'
+import { AddressBook } from '@chainlink/gauntlet-core'
 import { assertions } from '@chainlink/gauntlet-core/dist/utils'
 
 type COLOR = 'red' | 'green' | 'blue' | 'yellow' | 'cyan' | 'magenta'
```
RodrigoAD: The `TerraLogger` can be replaced by the new `Logger` from core right?

akhilchainani: yeah there's another PR open to address that - I still need to update that.

RodrigoAD: Let's include both changes in this PR

akhilchainani: cool I just rebased it with the new changes! So this should be good now @RodrigoAD

RodrigoAD (approved):
Looks good!

PR #267, opened 2022-04-20
Cache contract config and only check periodically the latest config block number

topliceanu:
This PR introduces a change to the way we fetch the contract's config from onchain logs. 
We now cache the result and only check the `latest_config_details` contract state periodically. If that doesn't show any new config, we serve the cached version. If there is a new config, we fetch that, cache it and return it from the Source.


pedrovsg (commented):
Where is the periodically part?

topliceanu:
Well, the Source's `Fetch()` method is called periodically by the monitor, I believe it's every 30s atm. 
What I did is to change the implementation of Fetch to serve the config from cache if it detects no update onchain, instead of blindly going to the chain all the time!

pedrovsg:
I got it now. You are getting the config from a client call always and only checking for logs if the config changes. My confusion was because of the word cache.

pedrovsg:
I got it now. You are getting the config from a client call always and only checking for logs if the config changes. My confusion was because of the word cache.

pedrovsg (approved):

PR #272, opened 2022-04-27
combine common test run workflows into common action

tateexon:
Tests are now passing.


.github/actions/build_test_artifacts/action.yml
```diff
@@ -0,0 +1,24 @@
+name: Build Test Artifacts
```
krebernisak: On Terra, we have `Build Test Contracts` while here it's called `Build Test Artifacts`. Let's try to keep more consistency between the two.

Is this only for testing, or could we reuse the action for `.github/workflows/contracts-release.yml` workflow?

.github/actions/build_test_artifacts/action.yml
```diff
@@ -0,0 +1,24 @@
...
+    - name: Cache cargo registry
+      uses: actions/cache@v2.1.7
+      with:
+        path: ~/.cargo/registry
+        key: ${{ runner.os }}-v2-cargo-registry-${{ hashFiles('**/Cargo.lock') }}
+    - name: Cache cargo index
+      uses: actions/cache@v2.1.7
+      with:
+        path: ~/.cargo/git
+        key: ${{ runner.os }}-v2-cargo-index-${{ hashFiles('**/Cargo.lock') }}
```
krebernisak: Cashing is not part of Is this `.github/workflows/contracts-release.yml` workflow currently, but perhaps we could add it and reuse the action?

.github/workflows/chaos.yml
```diff
@@ -27,99 +27,35 @@ jobs:
...
-        uses: actions/upload-artifact@v2.2.4
-        with:
-          name: test-logs
-          path: /home/runner/work/chainlink-terra/chainlink-terra/tests/e2e/logs
+        uses: smartcontractkit/ctf-ci-e2e-action@v1.0.0
+        with:
+          test_command_to_run: make test_chaos
+          cl_repo: 795953128386.dkr.ecr.us-west-2.amazonaws.com/chainlink
+          cl_image_tag: custom.${{ github.sha }}
+          download_artifacts_path: artifacts
```
krebernisak: Consider renaming this property to be more clear those are contract artifacts. There are also other types of artifacts we could test, like for example the Gauntlet binary.

.github/workflows/e2e_custom_cl.yml
```diff
@@ -48,78 +33,29 @@ jobs:
...
   e2e_cutsom_run_smoke_tests:
     name: E2E Custom Run Smoke Tests
     runs-on: ubuntu-latest
     needs: [e2e_custom_build_artifacts, e2e_custom_build_custom_chainlink_image]
     env:
       CGO_ENABLED: 1
     steps:
-      - name: Checkout the repo
-        uses: actions/checkout@f25a3a9f25bd5f4c5d77189cab02ff357b5aedeb # v2.4.1
-      - name: Install Nix
```
krebernisak, line 60: So all of these steps are now removed and there are implicitly run where?

This is now replaced with the `smartcontractkit/ctf-ci-e2e-action@v1.0.0` action which I assume takes care of some of these steps, but what about this repo custom steps like preparing the NIX env and running "Generate gauntlet executable", where is that run with the new flow?

tateexon: The ctf-ci-e2e-action (https://github.com/smartcontractkit/ctf-ci-e2e-action/blob/main/action.yml) takes care of preparing go, generating the gauntlet executable, and running the test. I am not adding the use of nix at this time since nix is only setup properly in 1 repo out of 6 or 7 that will be running this. We want our e2e runs to behave the same everywhere and that means nix is not the current solution. If we want nix then we need a more company wide push for its usage as a standard.

.github/actions/build_test_artifacts/action.yml
tateexon: Can do. I was only trying to clean up the e2e files so I didn't go look in all the others. Will take a look at the contracts-release file.

.github/workflows/chaos.yml
tateexon: That makes sense, I will update the ctf action

krebernisak (approved):
This looks good! Nice cleanup 🙌 

Before merging, please test the contract release action (push tag, delete tag) to see if it works as expected. Thx!

PR #273, opened 2022-04-27
Add secret to accept proposal step in initialize flow

kylesmartin:


RodrigoAD (approved):
Thanks @kylesmartin ! @krebernisak for viz

PR #277, opened 2022-05-03
Add gauntlet release workflow

kylesmartin:


lawrencexia:
Going to revisit the approach here. Keep the branch for reference for now

PR #278, opened 2022-05-04
Add inspect operators command

RodrigoAD:
Inspect command for operators:
- Fetch transmitters from aggregator contract
- Fetch and prompt LUNA balances for each transmitter
- Fetch last 100 transmissions. Prompt last transmission of each transmitter. Inactive transmitters are prompted


kylesmartin (approved):
Tested everything on my end and it looks good

PR #287, opened 2022-05-12
Median provider

connorwstein:


pkg/terra/relay.go
```diff
@@ -97,62 +88,39 @@ func (r *Relayer) Healthy() error {
...
-	if err != nil {
-		return nil, err
-	}
-	chainReader, err := chain.Reader(spec.NodeName)
-	if err != nil {
-		return nil, err
-	}
-	msgEnqueuer := chain.TxManager()
+func (r *Relayer) NewConfigWatcher(args relaytypes.ConfigWatcherArgs) (relaytypes.ConfigWatcher, error) {
+	return newConfigWatcher(r.ctx, r.lggr, r.chainSet, args)
```
jmank88: Might be worth being defensive here by never returning the typed nil `(*configWatcher)(nil)`:
```suggestion
	configWatcher, err := newConfigWatcher(r.ctx, r.lggr, r.chainSet, args)
	if err != nil {
		return nil, err
	}
	return configWatcher, nil
```

go.mod
```diff
@@ -7,8 +7,7 @@ require (
 	github.com/pelletier/go-toml v1.9.4
 	github.com/pkg/errors v0.9.1
 	github.com/prometheus/client_golang v1.12.1
-	github.com/satori/go.uuid v1.2.0
-	github.com/smartcontractkit/chainlink v1.3.0
+	github.com/smartcontractkit/chainlink v1.4.1-rc1.0.20220513135156-8d71a787718c
```
aalu1418: similar concern to solana changes (would like to have this referencing a develop commit at the very least), probably would have to happen in a follow up commit

connorwstein: yah follow up PRs required

jmank88 (dismissed):

aalu1418 (dismissed):

jmank88 (approved):

PR #289, opened 2022-05-13
use relay logger instead of core

jmank88:
https://app.shortcut.com/chainlinklabs/story/40148/chainlink-relay-create-logger-impl-for-testing-and-scripts
Requires
- [x] https://github.com/smartcontractkit/chainlink-relay/pull/49


pkg/terra/client/gas_price_estimator_test.go
```diff
@@ -99,7 +102,7 @@ func TestGasPriceEstimators(t *testing.T) {
 			"uluna": sdk.NewDecCoinFromDec("uluna", sdk.MustNewDecFromStr("10")),
 		})
 		gpe := NewMustGasPriceEstimator([]GasPricesEstimator{cachingFCD, gpeFixed}, lggr)
-		lggr.On("Warnf", mock.Anything, mock.Anything, mock.Anything).Twice()
+		t.Cleanup(assertLogsLen(t, 1))
```
jmank88, line 105: This one was incorrect, but these weren't actually being checked before because the `lggr.AssertExpectations(t)` at the end was missing.

go.mod
```diff
@@ -9,14 +9,15 @@ require (
...
-	github.com/smartcontractkit/chainlink-relay v0.0.0-20220422131434-0cef40354ecc
+	github.com/smartcontractkit/chainlink-relay v0.0.0-20220514135708-91ac8dd2d069
 	github.com/smartcontractkit/libocr v0.0.0-20220414173908-cdfa6bef133a
 	github.com/smartcontractkit/terra.go v1.0.3-0.20220108002221-62b39252ee16
 	github.com/stretchr/testify v1.7.1
 	github.com/tendermint/tendermint v0.34.15
 	github.com/terra-money/core v0.5.14
 	go.uber.org/multierr v1.8.0
 	go.uber.org/ratelimit v0.2.0
+	go.uber.org/zap v1.21.0
```
topliceanu, line 20: It would be cool is this wasn't a direct dependency. It's only needed for the log levels. I don't feel strongly about it, but you could alias the levels in `chainlink-relay/pkg/logger` to prevent this. Wdyt?

topliceanu (dismissed):

go.mod
jmank88: I *think* some imports would still leak through because of the observed logs...maybe not... :thinking:

topliceanu (approved):

Issue #294, opened 2022-05-23
replace `terrad` binary usage (previous releases removed)


aalu1418:
Related PR in core: https://github.com/smartcontractkit/chainlink/pull/6660

As of 23 May 2022, previous releases of terra-core have been removed from github - which also removes the referenced `terrad` version in core.

Potential solution: [build from source](https://github.com/terra-money/core#installation)

PR #295, opened 2022-05-27
remove chaos/soak cron runs for Terra

skudasov:
Removes chaos/soak pipelines to save some resources.


jmank88 (approved):

aalu1418 (approved):
🎉 

(I wonder if the tests are failing because `terrad` no longer exists on the terra repo? - issue is detailed here: #294 )

PR #309, opened 2022-10-28
Cosmwasm 1.0

archseer:


tests/e2e/gauntlet_deployer.go
```diff
@@ -41,7 +41,8 @@ type InspectionResult struct {
 }
 
 // GetDefaultGauntletConfig gets the default config gauntlet will need to start making commands
-// 	against the environment
+//
+//	against the environment
```
jmank88, line 45: Is this an incidental reformat due to the extra leading spaces?
```suggestion
// against the environment
```
Same below, although that one still makes sense :shrug:

pkg/terra/client/test_helpers.go
```diff
@@ -45,7 +44,7 @@ var minGasPrice = msg.NewDecCoinFromDec("uluna", msg.NewDecWithPrec(1, 3))
 // SetupLocalTerraNode sets up a local terra node via terrad, and returns pre-funded accounts, the test directory, and the url.
 func SetupLocalTerraNode(t *testing.T, chainID string) ([]Account, string, string) {
 	t.Skip("depends on terrad")
-	testdir, err := ioutil.TempDir("", "integration-test")
+	testdir, err := os.MkdirTemp("", "integration-test")
```
jmank88, line 47: Could go a step further here with the built in `t.TempDir()`, which is also cleaned up automatically too.
```suggestion
	testdir := t.TempDir()
```

go.work
```diff
@@ -0,0 +1,7 @@
+go 1.19
+
+use ./
+
+use ./ops/
+
+use ./tests/
```
jmank88, line 7: Did you mean to include `go.work`?

tests/e2e/gauntlet_deployer.go
archseer: I ran it through the `gopls` formatter and that's what I got 🤷🏻

go.work
archseer: Isn't it meant to be commited to the repository so the packages belong to the same [multi-module workspace](https://go.dev/doc/tutorial/workspaces)? That way modules in the tree depend on the same version without a whole bunch of back and forth bump commits.

(See https://github.com/smartcontractkit/chainlink-starknet/pull/174/commits where it took me 6 commits to do a simple dependency bump)

jmank88: IIUC it is only meant as a development aid.

https://dev.to/gophers/what-are-go-workspaces-and-how-do-i-use-them-1643

> **Best practises and Warnings**
Early posts from the community are suggesting that not committing your go.work file is preferable. This makes sense, it is used for local development and after you have made and tested your changes to a module, you should push it, tag it and then reference it in your go.mod file as usual. Therefore, sharing go.work files doesn't make sense.
>
> Consider adding go.work to your .gitignore.

jmank88: https://go.googlesource.com/proposal/+/master/design/45713-workspace.md

> **Preventing go.work files from being checked in to repositories**
go.work files that checked into repositories would cause confusion for Go users because they change the build configuration without the user explicitly opting in. Because of this they should be strongly discouraged. Though it's not clear that the Go tool should enforce this, other tools that vet repositories and releases should output warnings or errors for repositories containing go.work files. There may also be other mechanisms not yet considered in this document to discourage checked-in go.work files.

jmank88: I don't have an alternative solution, but I thought that we would need to start tagging the repo with differently formatted tags for something like this, specific to each module

jmank88: Ah, here: https://go.dev/doc/modules/managing-source#multiple-module-source

> Sourcing module code in subdirectories changes the form of the version tag you must use when publishing a module. You must prefix the version number part of the tag with the name of the subdirectory that is the module root.

> - Module path: example.com/mymodules/module1
> - Version tag: module1/v1.2.3

archseer:
Moving into #310

PR #317, opened 2023-03-15
TT-236 and TT-237 Cosmos helm chart and test skeleton

tateexon:


integration-tests/common/common.go
```diff
@@ -1,349 +1,127 @@
...
-	csaKeys, _, err := node.ReadCSAKeys()
-	if err != nil {
-		return "", err
-	}
-	shortNodeAddr := csaKeys.Data[0].Attributes.PublicKey[2:12]
-	shortOCRAddr := ocr2Addr[2:12]
-	return strings.ToLower(fmt.Sprintf("node_%s_contract_%s", shortNodeAddr, shortOCRAddr)), nil
+func (c *Common) Default(t *testing.T) {
+	c.K8Config = &environment.Config{NamespacePrefix: "chainlink-ocr-cosmos", TTL: c.TTL, Test: t}
+	// These can be uncommented when toml configuration is supposrted for cosmos in the chainlink node
```
tateexon, line 95: This will need to be fixed since the chainlink node currently does not accept the toml for cosmos.

integration-tests/common/common.go
```diff
@@ -1,349 +1,127 @@
 package common
```
tateexon, line 1: I didn't bring everything over from starknet for this file. Chances are a decent amount can be brought over mostly as is when writing the test.

integration-tests/common/test_common.go
```diff
@@ -0,0 +1,93 @@
+package common
```
tateexon, line 1: I didn't bring everything over from starknet for this file. Chances are a decent amount can be brought over mostly as is when writing the test.

integration-tests/smoke/ocr2_test.go
```diff
@@ -1,39 +1,32 @@
...
+)
 
-	AfterEach(func() {
-		By("Tearing down the environment", func() {
-			err := actions.TeardownSuite(state.Env, "logs", state.Nodes, nil, nil)
-			Expect(err).ShouldNot(HaveOccurred())
-		})
-	})
-})
+func TestOCRBasic(t *testing.T) {
```
tateexon, line 17: I only brought in the start of the test to spin up the environment for now until the rest of the test is able to be done.

ops/charts/devnet/templates/wasmd-deployment.yaml
```diff
@@ -0,0 +1,65 @@
...
+              value: cosm1234
+            - name: CHAIN_ID
+              value: testing
+            # If we need to change these for better control in testing we can
+            # - name: MONIKER
+            #   value: node001
+            # - name: STAKE
+            #   value: ustake
+            # - name: FEE
+            #   value: ucosm
```
tateexon, line 33: Not sure if `testing` is the needed chain id but it is what is set by default by the script. We can change more of this here as well as the other ones I left commented out.

ops/gauntlet/gauntlet_cosmos.go
```diff
@@ -0,0 +1,212 @@
+package gauntlet
```
tateexon, line 1: This file isn't used yet and I just brought it over from starknet. It will need to be updated to work with the cosmos gauntlet calls.

ops/wasmd/wasmd.go
```diff
@@ -0,0 +1,119 @@
+package wasmd
```
tateexon, line 1: This file will need to be made to work with cosmos instead of starknet as well.

integration-tests/common/common.go
archseer: We can build against https://github.com/smartcontractkit/chainlink/pull/8567 which will be merged into core soon

integration-tests/common/common.go
archseer: I wonder if there's more code we can abstract away and share between chains, `common.go`/`test_common.go` etc have quite a lot of code that looks mostly the same in each integration

tateexon: Oh we definitely can. I just wasn't worried about it right now since it will likely all get rewritten again for the BIF work with test templates and common libraries.

archseer (approved):

cfal (approved):

PR #319, opened 2023-03-17
rename FCD to LCD

cfal:


archseer:
FCD (full client) should be renamed to RPC, LCD means light client/read only.

PR #321, opened 2023-03-23
client/test_helpers: fix SetupLocalCosmosNode

### LINKED ISSUE #299: fix `terrad` command in pkg/solana/client
Disabled here: https://github.com/smartcontractkit/chainlink-terra/pull/298#issuecomment-1152747826

`terrad` is pointed to a newer version which has some changes

cfal:
BCI-978


wasmd.nix
```diff
@@ -0,0 +1,28 @@
...
+  sourceRoot = ".";
+
+  buildInputs = with pkgs; [ gnumake git go_1_20 which openssl cacert ];
+
+  buildPhase = ''
+    export HOME=$out
+    export GOSUMDB=off
+    export GOPROXY=direct
+    cd $src && make install
+    ln -s go/bin $out/bin
```
archseer: This should be under `installPhase`

wasmd.nix
```diff
@@ -0,0 +1,28 @@
...
+  version = "0.3.0";
+
+  src = fetchFromGitHub {
+    owner = "CosmWasm";
+    repo = "wasmd";
+    rev = "d48bb69b391edf92228708008dcd5bbe47b83128";
+    sha256 = "07nprps0da81krpl2a66qn823ybzghnx55zw7504wpynhgfa0bll";
+  };
+
+  sourceRoot = ".";
```
archseer: I think if you remove this then no need for `cd src`

cfal: for some reason, i still needed "cd src", even without `sourceRoot`

archseer (dismissed):

PR #324, opened 2023-03-31
pkg/cosmos: consolidate cosmos-sdk initialization, make configurable

cfal:


pkg/cosmos/params/params.go
```diff
@@ -0,0 +1,100 @@
...
+		Marshaler:         marshaler,
+		TxConfig:          txCfg,
+		Amino:             amino,
+	}
+}
+
+// TODO: import as params.MakeEncoding config
+var config = makeEncodingConfig()
+
+func SetupCosmosSdk(bech32Prefix, token string) {
```
archseer: Let's call it `Init`/`Initialize` rather than `Setup` since it shouldn't be called more than once

pkg/cosmos/params/params.go
```diff
@@ -0,0 +1,100 @@
...
+	sdkConfig.SetBech32PrefixForConsensusNode(bech32PrefixConsAddr, bech32PrefixConsPub)
+	sdkConfig.Seal()
+
+	for _, d := range []struct {
+		denom    string
+		decimals int64
+	}{
+		{token, 0},
+		{"m" + token, 3},
+		{"u" + token, 6},
```
archseer, line 81: Seems reasonable for now but we might run into different configs in the future/per token. I'm not sure all tokens use `0` decimals

cfal: yeah, was thinking about having an argument to pass in the whole map.. but i checked a few others and it seemed like the naming convention at least holds (eg uosmo). will double check on the decimals

cfal: i'm going to leave this for now as it seems like they're all pretty much the same:
https://github.com/search?q=RegisterDenom&type=code
https://github.com/osmosis-labs/osmosis/blob/8ec7dadbaf698a840034f43f38840c4ce3dd9599/app/params/config.go#L41

also, i could be misreading but it seems like the `0`, `3` and `6`, aren't specifically for number of decimals, the precision across the board is 18 - https://github.com/cosmos/cosmos-sdk/blob/47f46643affd7ec7978329c42bac47275ac7e1cc/types/decimal.go#L23

but for the exponent
https://github.com/cosmos/cosmos-sdk/blob/47f46643affd7ec7978329c42bac47275ac7e1cc/types/denom_internal_test.go#L13

so i think the "base" token having an exponent of 0 should be pretty standard, but could def see the other units changing for some chains... (although maybe milli, micro, nano, etc are also pretty standard and chances are low?)

i also went ahead and added `n` because its usage in `denom_internal_test.go` https://github.com/cosmos/cosmos-sdk/blob/47f46643affd7ec7978329c42bac47275ac7e1cc/types/denom_internal_test.go#L120

archseer (dismissed):

PR #341, opened 2023-07-12
packages-ts: rename to cosmos

cfal:


calvwang9 (dismissed):

packages-ts/gauntlet-cosmos-contracts/README.md
```diff
@@ -37,23 +37,23 @@ For our Access Controller contract, we could perform:
 
 - Deploy chainlink contract
 ```
-./bin/gauntlet-terra-macos access_controller:deploy --network=bombay-testnet
+./bin/gauntlet-cosmos-macos access_controller:deploy --network=bombay-testnet
 ```
-This will give us the contract address (`terra234`)
+This will give us the contract address (`cosmos234`)
```
calvwang9, line 42: curious if this is still the case?

packages-ts/gauntlet-cosmos-contracts/networks/.env.local
```diff
@@ -0,0 +1,6 @@
+NODE_URL=http://localhost:26657
+DEFAULT_GAS_PRICE="0.025ucosm"
```
calvwang9, line 2: should this be a decimal without the token denomination rather than a string, as in line with the other env files?

calvwang9 (dismissed):

packages-ts/gauntlet-cosmos-contracts/README.md
cfal: i can't speak on whether this works now, yet 😅 but, it needs to be the case so that the next steps can use the contract address

packages-ts/gauntlet-cosmos-contracts/networks/.env.local
cfal: seems like it shouldn't - and the mainnet file is incorrect, it requires at least a single alphabetical character after the number:

https://github.com/smartcontractkit/chainlink-cosmos/blob/develop/packages-ts/gauntlet-terra/src/commands/middlewares.ts#L49
https://github.com/cosmos/cosmjs/blob/v0.31.0/packages/stargate/src/fee.ts#L31

.. i updated the mainnet file to 0.5uatom for now.

cfal:
i'm merging this in now - the smoke test is still failing but that doesn't seem related to this afaict

PR #344, opened 2023-07-17
update integration tests

cfal:


cfal:
merging this in now to unblock other work dependent on RPC and gauntlet changes, and to prevent further conflicts.

PR #346, opened 2023-07-19
Update cosmos-sdk version

calvwang9:
https://smartcontract-it.atlassian.net/browse/BCI-1007

Updates the Cosmos SDK to 0.47, fixes breaking changes and deprecations, and also includes some minor formatting cleanup. Summary of main changes:
- Dependencies: 
  - cosmos-sdk 0.45.11 -> 0.47.4
  - wasmd 0.30 -> 0.40.1
  - tendermint -> cometbft 0.37.2
  - other minor updates
- Various renames across files due to deprecations
- Linting on imports and unused fields
- Broadcast mode 'block' was deprecated in v0.47, so need to use 'sync' mode instead and wait for the tx to be committed (added test helper fn)
- Updated some wasmd cli commands
- Makefile: rename relayer dir to pkg as it no longer exists, added `lint-go`


integration-tests/common/test_common.go
```diff
@@ -32,8 +32,8 @@ type Test struct {
 type ChainlinkClient struct {
 	NKeys          []client.NodeKeysBundle
 	ChainlinkNodes []*client.Chainlink
-	bTypeAttr      *client.BridgeTypeAttributes
-	bootstrapPeers []client.P2PData
```
calvwang9: linter complained that these were unused, can be uncommented when required

pkg/cosmos/client/test_helpers.go
```diff
@@ -205,3 +214,16 @@ func mustRandomPort() int {
...
+	for i := 0; i < 9; i++ { // max poll attempts to wait for tx commitment
+		txReceipt, err := tc.Tx(txHash)
+		if err == nil {
+			return txReceipt, true
+		}
+		t.Logf("tx not committed yet (attempt %d)", i+1)
+		time.Sleep(time.Second * 1) // TODO: configure dynamically based on block times
+	}
+	return nil, false
+}
```
calvwang9: Avg observed tx confirmation time was ~5 seconds, this polls for up to 10 seconds (arbitrary). We could set this to the default timeout defined in the Client (30 seconds) however in a testing environment 10 seconds should be sufficient, and increasing the wait duration may increase the duration of tests even more - currently the go unit tests with race detector enabled run for around 560-580 seconds, almost hitting the timeout at 600 seconds.

integration-tests/common/test_common.go
cfal: is it possible to leave `integration-test` changes for now? it'd save me from having to rebase and resolve conflicts for each commit (`git checkout origin/develop integration-tests` should reset it)

scripts/wasmd.sh
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
 container_name="chainlink-cosmos.wasmd"
-container_version="v0.31.0"
+container_version="v0.40.1"
```
cfal, line 4: can you cherry pick this commit and check that the container still starts up successfully? i'll remove it from the itests PR [`cc37e02` (#344)](https://github.com/smartcontractkit/chainlink-cosmos/pull/344/commits/cc37e02604e50ed7c1ee7d7a695c7eb4d57f4f4b)

cfal (commented):
can you check out the failing CI action? it looks possibly related: https://github.com/smartcontractkit/chainlink-cosmos/actions/runs/5641454331/job/15279538660?pr=346

calvwang9:
> can you check out the failing CI action? it looks possibly related: https://github.com/smartcontractkit/chainlink-cosmos/actions/runs/5641454331/job/15279538660?pr=346

@cfal yep I've seen this while working with the core repo, will have to fix in a follow up pr after this one is merged

scripts/wasmd.sh
calvwang9: yep looks like it starts just fine 👍

cfal (approved):

augustbleeds (approved):

PR #350, opened 2023-08-10
[BCI-1635] Integrate Generic Fixed Fee Estimator

yongkangc:
Context:
- To reduce code duplication and make use of the core node fee estimation logic, we are integrating fee estimator with cosmos txm as the first step to making use of the generalised platform. 

Changes:
- Bump Chainlink relay version to make use of fixed fee estimator
- Created getter and setter, as we might want to make use of mutexes to prevent race condition in the future for gas prices
- Integrated generic fixed fee estimator + add tests

Notes:
- PR that this is based off on
    - https://github.com/smartcontractkit/chainlink-relay/pull/149
    - https://github.com/smartcontractkit/chainlink/pull/10003
- This PR does not come with Retry mechanism. That would come with the integration of Confirmer from generic TXM platform.
-  next step is to fix the integration tests on core!
    
Ticket: 
- https://smartcontract-it.atlassian.net/browse/BCI-953
- https://smartcontract-it.atlassian.net/browse/BCI-1635


augustbleeds:
makes sense to me .. next step is to integrate the new Fixed Fee Estimator implementation to core ?

yongkangc:
> makes sense to me .. next step is to integrate the new Fixed Fee Estimator implementation to core ?

@augustbleeds That is right, the next step is to fix the integration tests on core!

jmank88 (dismissed):

pkg/cosmos/client/gas_price_estimator.go
```diff
@@ -17,16 +19,77 @@ var _ GasPricesEstimator = (*FixedGasPriceEstimator)(nil)
...
+func (gpe *FixedGasPriceEstimator) SetGasPrice(coin string, price sdk.DecCoin) {
+	gpe.gasPrices[coin] = price
+}
+
+// CalculateGasPrice calculates the minimum of the maximum input gas prices.
+// Parameters:
+// - maxGasPrice: max gas price set by the user or operator
+// - defaultGasPrice: default gas price
+// - maxGasPriceConfigured: max gas price originally configured for the node
+func (gpe *FixedGasPriceEstimator) CalculateGasPrice(
```
cfal: where does this get called? i can't seem to find anywhere besides gas_price_estimator_test.go

pkg/cosmos/client/gas_price_estimator.go
```diff
@@ -17,16 +19,77 @@ var _ GasPricesEstimator = (*FixedGasPriceEstimator)(nil)
...
 
 func (gpe *FixedGasPriceEstimator) GasPrices() (map[string]sdk.DecCoin, error) {
 	return gpe.gasPrices, nil
 }
 
+func (gpe *FixedGasPriceEstimator) GasPrice(coin string) sdk.DecCoin {
+	return gpe.gasPrices[coin]
+}
+
+func (gpe *FixedGasPriceEstimator) SetGasPrice(coin string, price sdk.DecCoin) {
```
cfal: - same for SetGasPrice, does this/will this get called anywhere besides the test?
- should it test that the coin is already one of the existing coins provided in the initial gas price map?

cfal (commented):
looks great overall! left some comments! also curious - does this gas estimator now conform to some generic gas estimator interface?

pkg/cosmos/client/gas_price_estimator.go
```diff
@@ -17,16 +19,77 @@ var _ GasPricesEstimator = (*FixedGasPriceEstimator)(nil)
...
+	maxGasPrice,
+	defaultGasPrice,
+	maxGasPriceConfigured sdk.DecCoin,
+) sdk.DecCoin {
+	calculatedGasPrice := fee.CalculateFee(
+		maxGasPrice.Amount.BigInt(),
+		defaultGasPrice.Amount.BigInt(),
+		maxGasPriceConfigured.Amount.BigInt(),
+	)
+	return sdk.NewDecCoinFromDec(coin, sdk.NewDecFromBigIntWithPrec(calculatedGasPrice, 18))
```
calvwang9: not sure if this is an assumption across the codebase but is precision always 18 decimals?

pkg/cosmos/client/gas_price_estimator.go
```diff
@@ -17,16 +19,77 @@ var _ GasPricesEstimator = (*FixedGasPriceEstimator)(nil)
...
+}
+
+// BumpGasPrice calculates a new gas price by bumping the current gas price by a percentage.
+// Parameters:
+// - currentGasPrice: current gas price (could be after bumping)
```
calvwang9: > currentGasPrice: current gas price (could be after bumping)

Could we elaborate a little bit on the 'could be after bumping' - in which cases would you retrieve the current gas price before/after bumping the fee?

pkg/cosmos/client/gas_price_estimator.go
```diff
@@ -17,16 +19,77 @@ var _ GasPricesEstimator = (*FixedGasPriceEstimator)(nil)
...
+// - originalGasPrice: original gas price (before bumping)
+// - maxGasPrice: max gas price
+// - maxBumpPrice: max gas price that can be bumped to
```
calvwang9, line 46: I know that we are just passing these straight to fee.CalculateBumpedFee, but just curious if there is a use case where these values should differ?

go.mod
```diff
@@ -12,8 +12,8 @@ require (
 	github.com/pelletier/go-toml v1.9.5
 	github.com/prometheus/client_golang v1.15.0
 	github.com/shopspring/decimal v1.3.1
-	github.com/smartcontractkit/chainlink-relay v0.1.7-0.20230801204930-f46c3ccc7815
-	github.com/smartcontractkit/libocr v0.0.0-20230606215712-82b910bef5c1
+	github.com/smartcontractkit/chainlink-relay v0.1.7-0.20230810155736-fc449f75f4fe
```
jmank88: Is this up to date?

yongkangc:
> looks great overall! left some comments! also curious - does this gas estimator now conform to some generic gas estimator interface?

The design for txm platform right now is such that there is no need for generic gas estimator interface anymore. The main thing we would need is just a bumping function and calculation function. That would be called by the [confirmer](https://github.com/smartcontractkit/chainlink/blob/78d61135bb0cdddfa30e7e7e619053f4ab1566ed/common/txmgr/confirmer.go#L788), which calls the `TxAttemptBuilder` to bump the gas.

Because `TxAttemptBuilder` would be a chain specific implementation, there would not be a generic interface to conform to.

pkg/cosmos/client/gas_price_estimator.go
yongkangc: In this case, the precision is 18 decimal in `sdkmath.LegacyPrecision` . Let me use that instead for clarity

yongkangc: This is because the underlying big int is represented at `Precision` as a string, so calling `BigInt()` increases the original value by 18 decimal points.

I.e
`sdk.MustNewDecFromStr("0.008")`` -> does not save it as 0.008 in the underlying big int. It saves as `8000000000000000``

yongkangc: thanks for bringing this up! i fixed it to clarify that in code more clearly

go.mod
yongkangc: thanks for the note, will update this to the latest version.

Edit: updated

pkg/cosmos/client/gas_price_estimator.go
yongkangc: Thanks for clarifying this calvin. 
```
// - maxGasPrice: max gas price
// - maxBumpPrice: max gas price that can be bumped to
```
Logically they are the same where they are a cap on the gas price we are willing to pay. 

Usecase wise, i would think that they can be different in the sense where 
- Max bump price can be a default value set across the node. Max gas price can be an input that can be raised or decreased. This is how EVM is doing it now. 

Do you think we should use the same value for both?

pkg/cosmos/client/gas_price_estimator.go
yongkangc: sure thing! apologies for the lack of clarity. currentGasPrice just means the price before that round of bumping, but it could have been bumped before. Let me update the commment for clarity. Thank you :)

yongkangc: updated comment for clarity!

pkg/cosmos/client/gas_price_estimator.go
yongkangc:
```
// - maxGasPrice: max gas price set by the user or operator
// - defaultGasPrice: default gas price
// - maxGasPriceConfigured: max gas price originally configured for the node
func (gpe *FixedGasPriceEstimator) CalculateGasPrice(
```

Currently, this is not being called anywhere. initially i was thinking of not having this in. However i thought that this might be useful in the future if we would like to calculate the gas price based on parameters instead of hard coding the fixed price value of `FallbackGasPrice() sdk.Dec` in for fixed price estimator. 

So instead of using `FallbackGasPrice() sdk.Dec`, we can pass in maxGasPrice for each transaction, and maxGasPriceConfigured by the node along with `FallbackGasPrice() sdk.Dec`, to get a gas price?

This could be relevant in the case of cosmos gas when `min-gas-prices` is a parameter local to each full-node? 

Do you think this part would be useful, or would it be clearer to remove it @cfal ?

pkg/cosmos/client/gas_price_estimator.go
yongkangc: > same for SetGasPrice, does this/will this get called anywhere besides the test?

For SetGasPrice, this would be called to set the price of gas after it is being calculated. Initially i wanted to combine the calculate gas price function and combine it with set gas price function to have a combined function, but felt it was cleaner to have them seperated.

> should it test that the coin is already one of the existing coins provided in the initial gas price map?

Thanks for suggesting this error case. Will implement it, apologies for missing it.

pkg/cosmos/client/gas_price_estimator.go
calvwang9: Currently, it seems to me that there are quite a lot of inputs and context required to use this bumped gas fee calculator, which could possibly be broken down into helper functions or calculated outside of the function and handled by the consumer. These are more suggestions for the generic fee estimator interface rather than this cosmos PR so these suggestions are not actionable here nor blocking this PR 👍 

we need to provide both:
- current gas price of the network
- original gas price

which is aggregated into a base gas price by taking the max. Having instead one input of `base_gas_price` means that less context needs to be shared with this helper function, the input/outputs are clearer to the consumer, and the consumer has the flexibility to choose which base to use. godoc documentation and/or helper fns can detail the recommended approach which is `max(current, original)`.

we also need to provide:
- min gas price bump
- max gas price bump
- max gas price overall
- % to bump gas price

essentially defining a bounded range (min to max) and a target (% of base) for the new gas price. context outside of this information (i.e. max bump vs max overall) is tangential to the calculation itself, and by aggregating it into one function input we can improve the clarity of the function inputs/outputs and follow a more single-responsibility approach.

yongkangc: Thanks for the feedback Calvin, appreciate the breakdown. Let me see if I can improve this by abstracting away the function parameters. 

Essentially:

1. Reduce the number of parameters by simplifying the context.
2. Separate the concerns of the functions to ensure they follow the single responsibility principle.
3. Enhance the readability of the code.
4. Simplify error handling and messages.

pkg/cosmos/client/gas_price_estimator.go
```diff
@@ -17,16 +19,77 @@ var _ GasPricesEstimator = (*FixedGasPriceEstimator)(nil)
...
-	return &FixedGasPriceEstimator{gasPrices: prices}
+func NewFixedGasPriceEstimator(prices map[string]sdk.DecCoin, lggr logger.SugaredLogger) *FixedGasPriceEstimator {
+	return &FixedGasPriceEstimator{gasPrices: prices, lggr: lggr}
 }
 
 func (gpe *FixedGasPriceEstimator) GasPrices() (map[string]sdk.DecCoin, error) {
 	return gpe.gasPrices, nil
 }
 
+func (gpe *FixedGasPriceEstimator) GasPrice(coin string) sdk.DecCoin {
```
cfal: GasPrices above returns a `(map, error)` while this GasPrice function is infallible and returns an `sdk.DecCoin` with no error.
- should these two functions be more consistent?
- do we need this new function? vs the caller doing `GasPrices()[coin]` themselves
- does it make sense for this to return an error if the coin doesn't exist?

pkg/cosmos/client/gas_price_estimator.go
cfal: thanks! no need to apologize!

pkg/cosmos/client/gas_price_estimator.go
cfal: if this is unused - i think it would be fine for this and related logic to be left out of this PR, and we can implement if it's needed. same for SetGasPrice - or if that one is used only for testing, it could be an internal function?

pkg/cosmos/client/gas_price_estimator.go
cfal: @calvwang9 is the decimals configured as part of the InitCosmosSdk denom initialization? curious if `sdkmath.LegacyPrecision` is accurate and used (especially because of that "legacy" prefix..)

cfal (commented):
can you rebase onto the updated `develop` branch? the integration tests should now pass and exercise your code.

jmank88 (dismissed):

pkg/cosmos/client/gas_price_estimator.go
yongkangc: Thanks for bringing this up @cfal ! Agreed with your point about how 
1. This function should be more consistent and return `(map, error)` 

2. 
> do we need this new function? vs the caller doing GasPrices()[coin] themselves

Initially, I thought that the getter and setter method would be useful especially when multiple `goroutines` bumps up, and therefore we might need mutexes. However upon deeper analysis of `TxnAttemptBuilder` and `confirmer` logic, it looks like there would not be a race condition because each transaction would only have one thread bumping the gas. 

I still think that this function would be useful for handling error if the coin does not exist. Instead of the caller having to define the error everytime the error is being called.

pkg/cosmos/client/gas_price_estimator.go
yongkangc: that make sense, we can always implement them if needed afterwards. Thanks for the suggestion!

pkg/cosmos/client/gas_price_estimator.go
ExtremelySunnyYK: that make sense, we can always implement them if needed afterwards. Thanks for the suggestion!

calvwang9: InitCosmosSdk already makes assumptions on the provided token when registering denominations on the cosmos SDK - https://github.com/smartcontractkit/chainlink-cosmos/blob/490a1b4c9921b404ba5c5813694dcdbc600d68df/pkg/cosmos/params/params.go#L87
i.e. {m, u, n} prefixes corresponding to {3, 6, 9} decimals relative to the base token (0 decimals)
If I understand correctly, this means that the current implementation could support a token with > 9 precision but would likely expect it to be 18 precision as per the standard. Given this then I think we are fine with proceeding as is, and can revisit if we ever have to deal with a native token that doesn't conform to this (unlikely IMO)

go.mod
```diff
@@ -160,4 +160,7 @@ replace (
 	github.com/btcsuite/btcd => github.com/btcsuite/btcd v0.22.1
 	github.com/gogo/protobuf => github.com/regen-network/protobuf v1.3.3-alpha.regen.1
 	github.com/hashicorp/go-plugin => github.com/jmank88/go-plugin v0.0.0-20230604120638-7bb12ec27e75
+
+	// until merged upstream: https://github.com/mwitkow/grpc-proxy/pull/69
+	github.com/mwitkow/grpc-proxy => github.com/jmank88/grpc-proxy v0.0.0-20230731114312-86ed94c93231
```
yongkangc: fixes for integration test. refer to https://smartcontract-it.atlassian.net/browse/BCI-1813

ExtremelySunnyYK:
There were some fixes needed to resolve dependencies on core and relay.
https://smartcontract-it.atlassian.net/browse/BCI-1813

yongkangc:
In order to fix up dependency issues with integration tests, i had to 
- Bump Grpc version
- Bump chainlink-core v2
- Bump chainlink/integration-test
- Fix common/client for integration test

[Commit](https://github.com/smartcontractkit/chainlink-cosmos/pull/350/commits/a165707a81a311b58c84d77317bbdfed2cddcbdf)
[Ticket](https://smartcontract-it.atlassian.net/browse/BCI-1813) with reference to starknet

integration-tests/go.mod
```diff
@@ -3,20 +3,21 @@ module github.com/smartcontractkit/chainlink-cosmos/integration-tests
...
-	github.com/smartcontractkit/chainlink-env v0.32.8
-	github.com/smartcontractkit/chainlink-relay v0.1.7-0.20230801204930-f46c3ccc7815
-	github.com/smartcontractkit/chainlink-testing-framework v1.11.6
-	github.com/smartcontractkit/chainlink/integration-tests v0.0.0-20230610030307-281eedc27b1f
-	github.com/smartcontractkit/chainlink/v2 v2.2.0-mercury-20230608.0.20230610030307-281eedc27b1f
-	github.com/smartcontractkit/libocr v0.0.0-20230606215712-82b910bef5c1
+	github.com/smartcontractkit/chainlink-env v0.36.0
+	github.com/smartcontractkit/chainlink-relay v0.1.7-0.20230814221409-567a8f84b413
+	github.com/smartcontractkit/chainlink-testing-framework v1.15.1
+	// github.com/smartcontractkit/chainlink/integration-tests v0.0.0-20230610030307-281eedc27b1f
```
jmank88:
```suggestion
```

integration-tests/common/client.go
```diff
@@ -16,22 +16,28 @@ import (
 )
 
 type ChainlinkClient struct {
-	ChainlinkNodes []*client.Chainlink
+	ChainlinkNodes []*client.ChainlinkK8sClient
```
cfal: we should be able to initialize the ChainlinkClient directly without the K8s specific ConnectChainlinkNodes eg: https://github.com/smartcontractkit/chainlink/blob/8e89b62488d74953312e1aa94fb68b2c9cc21037/integration-tests/docker/test_env/cl_node.go#L250

and this field can be `[]*client.ChainlinkClient`, we don't use any of the k8s functionality

cfal (changes requested):

integration-tests/common/client.go
yongkangc: thanks for pointing this out and redirecting me to the `ChainlinkClient`! Appreciate it.

integration-tests/go.mod
yongkangc: apologies for missing this out. Fixed

integration-tests/common/client.go
ExtremelySunnyYK: I realised that ChainlinkClient does not have the method to `ConnectChainlinkNodes`, only `[ChainlinkK8sClient](https://github.com/smartcontractkit/chainlink/blob/cosmos-test-keys/integration-tests/client/chainlink_k8s.go#L77) has that. Is there an equivalent `ConnectChainlinkNodes`that we can use? ([location](https://github.com/smartcontractkit/chainlink-cosmos/pull/350/files#diff-adcbd79784f9beb82e87b4a73b78298506993ad29fe938f92b5f2c8582b20b53L27))

yongkangc: Created ChainlinkClient functions to get it to work: 

https://github.com/smartcontractkit/chainlink-cosmos/pull/350/commits/c98307eabcda3280688aa730b7b63d4314aa8db6

calvwang9 (dismissed):
nice work!

integration-tests/common/client.go
```diff
@@ -16,15 +18,15 @@ import (
...
+	ChainlinkNodes []*client.ChainlinkClient
 	NodeKeys       []client.NodeKeysBundle
 	bTypeAttr      *client.BridgeTypeAttributes
 	bootstrapPeers []client.P2PData
 }
 
 // CreateKeys Creates node keys and defines chain and nodes for each node
 func NewChainlinkClient(env *environment.Environment, nodeName string, chainId string, tendermintURL string) (*ChainlinkClient, error) {
-	nodes, err := client.ConnectChainlinkNodes(env)
+	nodes, err := ConnectChainlinkNodes(env)
```
cfal: can you add a TODO to remove `env`? we don't actually need `environment.Environment` anymore because we're using it to store a `ChainlinkNodeDetails` array: https://github.com/smartcontractkit/chainlink-cosmos/blob/29ae98077f2bdd42fff33ab4bdc66bf28783cdd7/integration-tests/common/common.go#L197

and then deconstructing that array of objects and using it in the new `ConnectChainlinkNodes`

cc @calvwang9

integration-tests/common/client.go
```diff
@@ -169,3 +171,34 @@ func (cc *ChainlinkClient) CreateJobsForContract(chainId, nodeName, p2pPort, moc
 	}
 	return nil
 }
+
+// ConnectChainlinkNodes creates a chainlink client for each node in the environment
+// This is a non k8s version of the function in chainlink_k8s.go
+// https://github.com/smartcontractkit/chainlink/blob/cosmos-test-keys/integration-tests/client/chainlink_k8s.go#L77
+func ConnectChainlinkNodes(e *environment.Environment) ([]*client.ChainlinkClient, error) {
```
cfal: nit: this can be a private function

integration-tests/common/client.go
yongkangc: sure thing! thanks for the note

pkg/cosmos/client/gas_price_estimator.go
cfal: makes sense, thanks!

cfal (dismissed):

cfal (dismissed):

go.mod
jmank88: You can use `github.com/smartcontractkit/grpc-proxy v0.0.0-20230731113816-f1be6620749f` now instead of my personal repo:
https://github.com/smartcontractkit/chainlink/blob/develop/go.mod#L387

jmank88 (dismissed):

go.mod
yongkangc: Thanks Jordan, will do that.

Edit: Bumped

cfal (approved):

PR #353, opened 2023-08-14
BCI-1749: Setup local env in integration tests

calvwang9:
[BCI-1749](https://smartcontract-it.atlassian.net/browse/BCI-1749)

Remove the need to manually run setup scripts before running integration tests. Also minor cleanup of scripts/files, and source CL node config from `integration-tests/common.go` instead of hardcoded in `core.sh`.

`CORE_IMAGE` env var (used in `core.sh`) is an optional env var to be defined when running integration tests.


integration-tests/common/common.go
```diff
@@ -122,18 +122,25 @@ Enabled = true
 ChainID = '%s'
 [[Cosmos.Nodes]]
 Name = 'primary'
-TendermintURL = '%s'
+TendermintURL = 'http://host.docker.internal:26657'
```
cfal: - it's strange that we have `nodeUrl` as configurable - but then this URL is hard-coded in the config. perhaps have an `INTERNAL_NODE_URL` env var this this set to the default, similar to `NODE_URL`?
- could we also update `defaultNodeUrl`? `http://tendermint-rpc:26657` doesn't work for the local env, maybe 127.0.0.1:26657 would be a more sane default

integration-tests/ocr2_test.go
```diff
@@ -35,6 +36,23 @@ func TestOCRBasic(t *testing.T) {
 	commonConfig := common.NewCommon(t)
 	commonConfig.SetLocalEnvironment()
 
+	logger.Info().Msg("Starting wasmd container...")
```
cfal: it'd be nice if this test could still support other environments - eg `commonConfig.SetK8sEnvironment`

perhaps we can move all this initialization into `SetLocalEnvironment`? and then any K8S specific code that we end up re-adding could be in `SetK8sEnvironment`, and a single function call change (perhaps governed by some future env var) would be enough to switch environments but run this same test.

scripts/core.sh
```diff
@@ -11,36 +11,17 @@ container_version="2.3.0"
...
+	echo "No CL_CONFIG env var provided." >&2
+	exit 1
+fi
 
+platform_arg=""
 if [ -n "${CORE_IMAGE:-}" ]; then
 	image_name="${CORE_IMAGE}"
 else
 	image_name="smartcontract/chainlink:${container_version}"
+	platform_arg="--platform linux/amd64"
```
cfal, line 24: nit: `--platform` doesn't need to be part of `platform_arg` and could be part of the command line, eg `--platform "${platform_arg}"`

integration-tests/ocr2_test.go
```diff
@@ -35,6 +36,23 @@ func TestOCRBasic(t *testing.T) {
...
+	logger.Info().Msg("Starting postgres container...")
+	err = exec.Command("../scripts/postgres.sh").Run()
+	require.NoError(t, err, "Could not start postgres container")
+	logger.Info().Msg("Starting mock adapter...")
+	err = exec.Command("../scripts/mock-adapter.sh").Run()
+	require.NoError(t, err, "Could not start mock adapter")
+	logger.Info().Msg("Starting core nodes...")
+	cmd := exec.Command("../scripts/core.sh")
+	cmd.Env = os.Environ()
+	cmd.Env = append(cmd.Env, fmt.Sprintf("CL_CONFIG=%s", commonConfig.ChainlinkConfig))
```
cfal: nit: the previous line could be removed and this could be `append(os.Environ(), ...)` directly

scripts/core.sh
calvwang9: If core image was provided then I wanted to omit the platform arg altogether as it doesn't seem to work in that case, hence I included the `--platform` as well. I tried this out but it errors - is there a better way to achieve this?

cfal: got it - i think the problem is that on Mac, it's not building an amd64 image by default, you'd have to specify something like `linux/arm64`. one solution is to provide the platform arg for custom images using:
```
platform_arg=$(docker image inspect "${image_name}" -f "{{.Os}}/{{.Architecture}}")
```

cfal: feel free to leave it and skip the argument though - agree that also seems reasonable

PR #355, opened 2023-08-15
BCI-1511: Gauntlet Test Link

augustbleeds:
*UPDATE: This PR has been enveloped into https://github.com/smartcontractkit/chainlink-cosmos/pull/364*

Most of the line code changes are from modifying the generated json schemas

Please filter out .json files to better review the PR

image

- [x] Write in-line github comments for easier code review

Couple of changes for this
* schema fixes to be able to use codegen for typescript clients
* fix schema scripts to use improved write_api macro to generate json api
* add typescript clients
* changes to gauntlet-cosmos to be able to support invocation via jest
* setup and teardown wasmd functionality + debug mode for quick local dev / testing 
* fix wasmd.nix and shell.nix by adding packages to support mac and ledger
* augment Github CI/CD to run gauntlet tests and cache the artifacts/ dir
* modify jest configs to support running just on yarn workspaces from the root folder


.github/workflows/gauntlet.yml
```diff
@@ -61,5 +61,18 @@ jobs:
         with:
           name: chainlink-cosmos
           authToken: '${{ secrets.CACHIX_AUTH_TOKEN }}'
+      - name: Cache Wasmd Artifacts
```
augustbleeds, line 64: cache the root dir "artifacts/" folder which contains the optimized .wasm bytecode files. This usually takes a while to compile

.gitignore
```diff
@@ -24,6 +24,8 @@ integration-tests/logs
 integration-tests/smoke/logs
 integration-tests/soak/logs
 tmp-manifest-*
-artifacts
+/artifacts
```
augustbleeds, line 27: only cache the root directory artifacts folder (we commit to source other artifacts/ folders for cw-plus contracts)

.gitignore
```diff
@@ -24,6 +24,8 @@ integration-tests/logs
...
 .local-mock-server
+codegen
+devAccounts.json
```
augustbleeds, line 30: json file that contains test accounts for local wasmd -- is created everytime local wasmd is setup in the script (see other code below)

.gitignore
```diff
@@ -24,6 +24,8 @@ integration-tests/logs
 integration-tests/smoke/logs
 integration-tests/soak/logs
 tmp-manifest-*
-artifacts
+/artifacts
 .local-mock-server
+codegen
```
augustbleeds, line 29: folder that is auto-generated by https://github.com/CosmWasm/ts-codegen

contracts/access-controller/Cargo.toml
```diff
@@ -22,6 +22,7 @@ backtraces = ["cosmwasm-std/backtraces"]
 library = []
 
 [dependencies]
+cosmwasm-schema = { version = "1.1.5" }
```
augustbleeds, line 25: we have to move cosmwasm-schema from dev-dependency to dependency to use the new write_api! macro

contracts/access-controller/examples/schema.rs
```diff
@@ -1,17 +1,18 @@
...
 fn main() {
     let mut out_dir = current_dir().unwrap();
     out_dir.push("schema");
     create_dir_all(&out_dir).unwrap();
-    remove_schemas(&out_dir).unwrap();
 
-    export_schema(&schema_for!(InstantiateMsg), &out_dir);
-    export_schema(&schema_for!(ExecuteMsg), &out_dir);
-    export_schema(&schema_for!(QueryMsg), &out_dir);
+    write_api! {
```
augustbleeds, line 13: Instead of individually, writing the schema for instantiate, execute, and query messages, the write_api! macro generates all of them + a centralized json file that contains all the json fields necessary to describe the entire contract.

contracts/access-controller/src/msg.rs
```diff
@@ -1,22 +1,25 @@
-use schemars::JsonSchema;
-use serde::{Deserialize, Serialize};
+use cosmwasm_schema::{cw_serde, QueryResponses};
+use cosmwasm_std::Addr;
 
-#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Eq, JsonSchema)]
+#[cw_serde]
```
augustbleeds: cw_serde macro includes all of the commonly used macros besides Eq

contracts/access-controller/src/msg.rs
```diff
@@ -1,22 +1,25 @@
...
     AddAccess { address: String },
     RemoveAccess { address: String },
     TransferOwnership { to: String },
     AcceptOwnership,
 }
 
-#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Eq, JsonSchema)]
-#[serde(rename_all = "snake_case")]
+#[cw_serde]
+#[derive(Eq, QueryResponses)]
```
augustbleeds: QueryResponses allows you to associate the return type with the query message. This is a must have for ts-codegen

contracts/deviation-flagging-validator/examples/schema.rs
```diff
@@ -1,19 +1,39 @@
...
+
+    let mut new_location = main_dir.clone();
+    new_location.push("deviation-flagging-validator.json");
+    rename(&main_file, &new_location).unwrap();
+
+    // put other hand-exported schemas in seperate folder
+    let mut other_dir = out_dir.clone();
+    other_dir.push("other");
+    create_dir_all(&other_dir).unwrap();
+    export_schema(&schema_for!(State), &other_dir);
```
augustbleeds, line 38: I don't know if we actually use this State schema but I kept it here in case we do. I'll remove it if I find it's unused after all gauntlet tests are finished

jest.config.js
```diff
@@ -23,5 +23,16 @@ module.exports = {
         },
       },
     },
+    {
```
augustbleeds, line 26: add gauntlet-cosmos-cw-plus to be discoverable as a jest project

packages-ts/gauntlet-cosmos-contracts/networks/.env.local
```diff
@@ -1,4 +1,4 @@
-NODE_URL=http://localhost:26657
```
augustbleeds, line 1: for some reason, wasmd does not support ipv6 so you have to specify the ipv4 address

packages-ts/gauntlet-cosmos-contracts/src/lib/contracts.ts
```diff
@@ -58,7 +58,9 @@ export abstract class Contract {
     if (version === 'local') {
       // Possible paths depending on how/where gauntlet is being executed
       const possibleContractPaths = [
+        path.join(__dirname, '../../../../artifacts'),
```
augustbleeds, line 61: make artifacts discoverable using __dirname

packages-ts/gauntlet-cosmos-contracts/src/lib/contracts.ts
```diff
@@ -95,19 +101,14 @@ export abstract class Contract {
     const abi = possibleContractPaths
       .filter((path) => existsSync(`${path}/${this.dirName}/schema`))
       .map((contractPath) => {
-        const toPath = (type) => {
```
augustbleeds, line 98: simply contract path logic

packages-ts/gauntlet-cosmos-contracts/src/scripts/codegen.js
```diff
@@ -0,0 +1,61 @@
+const codegen = require('@cosmwasm/ts-codegen').default
```
augustbleeds, line 1: codegen reads the json api from each path to generate the typed client

scripts/wasmd.sh
```diff
@@ -4,7 +4,7 @@ container_name="chainlink-cosmos.wasmd"
 container_version="v0.40.1"
 genesis_account="wasm1lsagfzrm4gz28he4wunt63sts5xzmczwda8vl6"
 
-set -euo pipefail
+set -euox pipefail
```
augustbleeds, line 7: log output

scripts/wasmd.sh
```diff
@@ -35,7 +35,8 @@ docker run \
 	--name "${container_name}" \
 	"cosmwasm/wasmd:${container_version}" \
 	"./setup_and_run.sh" \
-	"${genesis_account}"
+	"${genesis_account}" \
+	"$@"
```
augustbleeds: pass in more account addresses to be created

wasmd.nix
```diff
@@ -13,7 +13,7 @@ stdenv.mkDerivation rec {
     sha256 = "sha256-KBchSnIETM+HPngqHeajiqLYrhHhXY531C+Q/aSDzl4=";
   };
 
-  buildInputs = with pkgs; [ gnumake git go_1_20 which openssl cacert ];
+  buildInputs = with pkgs; [ gnumake git go_1_20 which openssl cacert gcc ];
```
augustbleeds, line 16: fix wasmd build on mac

.github/workflows/gauntlet.yml
```diff
@@ -61,5 +61,18 @@ jobs:
...
+        uses: actions/cache@v3
+        id: cache-wasmd-artifacts
+        env:
+          cache-name: cache-wasmd-artifacts
+        with:
+          path: artifacts
+          key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('contracts') }}
```
cfal, line 71: this is a great optimization, the wasmd artifacts build step is also pretty slow in some other workflows. i'm not sure this `key` is sufficient though - what if the Makefile or `packages-ts` changes? couldn't that affect artifacts output?
for example, some `.wasm` artifacts are directly downloaded in `packages-ts`: https://github.com/smartcontractkit/chainlink-cosmos/blob/861c937bfb8536e463fc3dc4dd43b57e0e4138b7/packages-ts/gauntlet-cosmos-contracts/src/lib/contracts.ts#L128

packages-ts/gauntlet-cosmos-contracts/test/utils.ts
```diff
@@ -0,0 +1,109 @@
...
+    const wallet = await DirectSecp256k1HdWallet.generate(12, { prefix: 'wasm' })
+    const account = await wallet.getAccounts()
+    return account[0]
+  })
+  let accounts = await Promise.all(otherAccounts)
+
+  const otherAddresses = accounts.map((a) => a.address).join(' ')
+  const wasmdScript = path.join(__dirname, '../../../scripts/wasmd.sh')
+
+  execSync(`${wasmdScript} ${otherAddresses}`)
```
cfal: i'm curious if we should be adding other addresses as validators. if these are meant to be normal addresses that only require some tokens (link/native) - could we just send it to them from the deployer address instead?

we actually do this in the integration test: https://github.com/smartcontractkit/chainlink-cosmos/blob/861c937bfb8536e463fc3dc4dd43b57e0e4138b7/integration-tests/ocr2_test.go#L65

but reconsidering that implementation - this would be nice to have as a gauntlet action to use here and in these tests as well

cc @calvwang9

contracts/access-controller/examples/schema.rs
cfal: can you move contracts/rust/non-gauntlet changes into a separate PR? 🙏

cfal (changes requested):
looks good, added some initial comments!

.github/workflows/gauntlet.yml
augustbleeds: This is just for the chainlink contracts that we write (since it is just calling build-contracts.sh at the end of the day). the cosmos base contracts will always be downloaded I believe the way things are written now

packages-ts/gauntlet-cosmos-contracts/test/utils.ts
augustbleeds: Yeah that's a good point, I think the original validator already has enough token as it is so I'll just create new accounts within the test

.github/workflows/gauntlet.yml
cfal: please double check, but i believe the caching step occurs at the end of your workflow - which means it would store everything that's been placed in the `artifacts/` path, and then restore everything, including downloaded wasm code. there is then a file exists check here which would stop a download of a new wasm file from happening: https://github.com/smartcontractkit/chainlink-cosmos/blob/861c937bfb8536e463fc3dc4dd43b57e0e4138b7/packages-ts/gauntlet-cosmos-contracts/src/lib/contracts.ts#L69

ideally, the caching should only cache the compiled contracts. perhaps it can be moved into a subdirectory/isolated from `artifacts/`.

augustbleeds: Here's what my root artifacts/ folder looks like after running local tests ~

image

I'm pretty confident that the cw-plus contracts are never stored in the artifacts/ folder in the default workflow because it is read as an array buffer.

The way it's written, I think the original author intended for cw-plus contracts to be downloaded (https://github.com/smartcontractkit/chainlink-cosmos/blob/861c937bfb8536e463fc3dc4dd43b57e0e4138b7/packages-ts/gauntlet-cosmos-contracts/src/lib/contracts.ts#L130) whereas chainlink contracts are found locally because cw-plus contracts never have the version set to 'local'

packages-ts/gauntlet-cosmos-contracts/test/command/link.test.ts
```diff
@@ -0,0 +1,120 @@
...
+  }[]
+}
+
+const ONE_TOKEN = BigInt('1000000000000000000')
+const MNEMONIC =
+  'surround miss nominee dream gap cross assault thank captain prosper drop duty group candy wealth weather scale put'
+const NODE_URL = 'http://127.0.0.1:26657'
+const DEFAULT_GAS_PRICE = '0.025ucosm'
+const NETWORK = 'local'
+const TIMEOUT = 90000
```
calvwang9, line 21: nit: not all of these consts are used and some are already defined in utils.ts

packages-ts/gauntlet-cosmos-contracts/test/utils.ts
```diff
@@ -0,0 +1,117 @@
...
+
+/**
+ * Initializes Wasmd and Contracts, unless SKIP_WASMD_SETUP=true
+ * Can save time for debugging / local testing if Wasmd and base contracts have been setup already
+ *
+ * Will read initialized account addresses from local file system if SKIP_WASMD_SETUP=true
+ *
+ * @returns {string[]} Initialized account addresses
+ */
+export const maybeInitWasmd = async () => {
```
calvwang9, line 44: nit: 'maybeInitWasmd' seems kinda confusing semantically (is wasmd initialized or not?)- perhaps just 'initWasmd' would suffice?

calvwang9 (approved):
looks great! just small nits from me

PR #359, opened 2023-08-21
BCI-1511: only contract schema changes

augustbleeds:
Originally diffed from https://github.com/smartcontractkit/chainlink-cosmos/pull/355 to just seperate out the schema changes

Changes:
* For query messages use serialized objects instead of enums (ex: `{ "owner": {} }` instead of `"owner"` to query the owner of a contract) for best practices + type generation compatibility
* Modify schema generation script to use new write_api! macro. It creates a main <contract>.json file which describes the entire contract json api and raw/ file folder which contains individual json schemas
* Add main <contract>.json cw-plus files in gauntlet-cw-plus package

TODO: make changes to gauntlet to support the query message types (in a later PR)


contracts/access-controller/Cargo.toml
```diff
@@ -22,6 +22,7 @@ backtraces = ["cosmwasm-std/backtraces"]
 library = []
 
 [dependencies]
+cosmwasm-schema = { version = "1.1.5" }
```
archseer, line 25: Hmm why the need for this? The crate's readme specifically states it's a dev dependency

augustbleeds: https://book.cosmwasm.com/basics/good-practices.html#json-schema -- just copied over here. It also didn't work as a dev dependency as I wasn't able to use it in schema.rs

archseer (approved):
LGTM, it does look like the schema conventions changed slightly (so no more dev dependency, custom macros)

PR #360, opened 2023-08-24
[BCI-1477] Implement cosmos specific txn type for generic txm

yongkangc:
This PR implements implements chain specific transaction type based on the generic Transaction manager transaction interface. 

**Context:**
- In order for cosmos to make use of generic txm logic, it requires to implement the `TxRequest` type and encoding/decoding methods are needed. 
- The EVM has a similar implementation here: https://github.com/smartcontractkit/chainlink/blob/7b57ad4c15a2286377ba8cb8775d6c7aa2796251/core/chains/evm/txmgr/models.go#L16-L17

**Changes:**
- bumping chainlink-relay to make use of the generic `Txn Types` in the PR https://github.com/smartcontractkit/chainlink-relay/pull/154
- implemented chain specific transaction types and encoding/decoding methods. 
- tests

**Note:** 
- Not sure if we would want to merge this, as this PR is dependent on Txn-types on chainlink-relay in the PR https://github.com/smartcontractkit/chainlink-relay/pull/154. However we are still in the middle of persistence layer proposal, and sqlx dependency has yet to be extracted out. See [comment](https://github.com/smartcontractkit/chainlink-relay/pull/154#issuecomment-1687322299) for the discussion. 
	- Might have to consult @aalu1418 for path forward. 

**References:**
- [Design docs](https://www.notion.so/chainlink/Design-Doc-Cosmos-TXM-cadeeccc982f4a2b9e801921a6f3a51f?pvs=4#5c5768a0c264467c8311a96256cee249)
- Tickets: https://smartcontract-it.atlassian.net/browse/BCI-1477


pkg/cosmos/txmgr/models.go
```diff
@@ -0,0 +1,107 @@
...
+		err := ms.Unmarshal(raw)
+		if err != nil {
+			return nil, err
+		}
+		return &ms, nil
+	}
+	return nil, errors.Errorf("unrecognized message type: %s", msgType)
+}
+
+// A wrapper for sdk.Address. Uses fixed max address size of 32-byte size to satisfy Hashable interface.
```
prashantkumar1982: Just curious, why do we use 32-bytes, and not just []bytes?
As in this comment, I don't think the Hashable interface enforces us to have [32]byte. 

Using just []byte would allow us to use sdk.Address directly.

yongkangc: we would need to use `32-bytes` because it needs to conform to the `comparable` constraint implemented by the hashable interface. Would update the comment to make this clearer

pkg/cosmos/txmgr/models.go
```diff
@@ -0,0 +1,107 @@
...
+// Create a new generic txmgr Address from a sdk.Address
+func NewAddress(addr sdk.AccAddress) Address {
+	var a Address
+	copy(a[:], addr[:])
+	return a
+}
+
+// Convert a generic txmgr Address to a Cosmos sdk.AccAddress
+func ToCosmosAddress(addr Address) sdk.AccAddress {
+	return sdk.AccAddress(bytes.TrimRight(addr[:], "\x00"))
```
prashantkumar1982, line 87: If I understand this correctly, does this remove all trailing 0s from the address?

What if the actual cosmos address had 0s in the address?
Or, I assume cosmos address wouldn't have 0s(unicode 0s, not string of "0") in the address, so its safe to do this?

yongkangc:
The Staging branch has been changed to `cosmos-generic-txm-staging`

pkg/cosmos/txmgr/models.go
yongkangc: > If I understand this correctly, does this remove all trailing 0s from the address?


yup, this will remove trailing zeros from the addresses. 

> What if the actual cosmos address had 0s in the address?

The cosmos addresses can have 0s in the address, but it would not have unicode 0s

Cosmos addresses are derived from public keys using the Bech32 encoding scheme. Because Bech32 uses a specific set of 32 characters, the Unicode null character (U+0000) would not be possible.

[Reference](https://river.com/learn/terms/b/bech32/#:~:text=The%20Bech32%20alphabet%20contains%2032,includes%20an%20error%20detection%20mechanism.): 
Bech32 is an encoding scheme used to encode SegWit addresses and Lightning invoices. The Bech32 alphabet contains 32 characters, including lowercase letters a-z and the numbers 0-9, excluding the number 1 and the letters ‘b’, ‘i’, ‘o’ to avoid reader confusion. Bech32 also includes an error detection mechanism.

PR #363, opened 2023-08-25
BCI-1796: Support dynamic adapter selection

calvwang9:
https://smartcontract-it.atlassian.net/browse/BCI-1796

This PR adds support for dynamically selecting the Cosmos adapter (currently only cosmos and injective) based on the bech32 prefix of the chain.

### DIFF
--- a/pkg/cosmos/relay.go
+++ b/pkg/cosmos/relay.go
@@ -16,18 +16,9 @@ import (
 	"github.com/smartcontractkit/chainlink-cosmos/pkg/cosmos/params"
 )
 
-type CosmosChainType string
-
 const (
-	Wasm      CosmosChainType = "wasm"
-	Injective CosmosChainType = "injective"
+	InjectivePrefix string = "inj"
 )
-
-var Bech32PrefixToCosmosChainType = map[string]CosmosChainType{
-	"cosmos": Wasm,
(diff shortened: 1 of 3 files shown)


pkg/cosmos/relay.go
```diff
@@ -12,9 +12,23 @@ import (
...
 )
 
+type CosmosChainType string
+
+const (
+	Wasm      CosmosChainType = "wasm"
+	Injective CosmosChainType = "injective"
+)
+
+var Bech32PrefixToCosmosChainType = map[string]CosmosChainType{
```
cfal: i think it could be easier to maintain to be either injective, or default to Wasm for any other unit. that way adding any other cosmwasm chain would strictly be a config change without requiring a PR.

cfal (commented):
looks good 🎉 left 1 comment on chain type selection

pkg/cosmos/relay.go
calvwang9: good idea, cleaned it up a bit 👍

cfal (approved):

calvwang9:
Integration tests currently failing due to a type mismatch from WIP refactor of chain sets - likely resolved by this [PR](https://github.com/smartcontractkit/chainlink/pull/10349)

PR #364, opened 2023-08-29
BCI-1511: more gauntlet tests

augustbleeds:
Note: I closed https://github.com/smartcontractkit/chainlink-cosmos/pull/355 and pulled the schema changes into https://github.com/smartcontractkit/chainlink-cosmos/pull/359

I decided to also use the lite mode on CI so I could reduce the total test time from ~13 minutes to ~5 minutes because you only need to start wasmd once. You can still test with yarn test:full if you'd like for a new wasmd container to startup at the beginning of every testfile

Changes:
- [x] add typescript clients
- [x] changes to gauntlet-cosmos to be able to support invocation via jest
- [x] setup and teardown wasmd functionality + debug mode for quick local dev / ci
- [x] fix wasmd.nix and shell.nix by adding packages to support mac and ledger
- [x] augment Github CI/CD to run gauntlet tests and cache the artifacts/ dir
- [x] modify jest configs to support running just on yarn workspaces from the root folder
- [x] add tests for link token, proxy, ocr2, access controller, flags, validator contracts

core ref: cosmos-keystore


contracts/deviation-flagging-validator/src/contract.rs
```diff
@@ -185,7 +185,7 @@ pub fn is_valid(flagging_threshold: u32, previous_answer: i128, answer: i128) ->
         return Ok(true);
     }
 
-    // https://github.com/rust-lang/rust/issues/89492
+    // TODO: incorporate fix from https://github.com/rust-lang/rust/issues/89492
```
augustbleeds, line 188: linked issue has been resolved so we are free to tackle this item now

contracts/flags/schema/flags.json
```diff
@@ -147,10 +147,10 @@
       {
         "type": "object",
         "required": [
-          "Owner"
+          "owner"
```
augustbleeds, line 150: change to snake case convention

packages-ts/gauntlet-cosmos-contracts/networks/.env.local
```diff
@@ -1,4 +1,4 @@
-NODE_URL=http://localhost:26657
+NODE_URL=http://127.0.0.1:26657
```
augustbleeds, line 1: added because fetch defaults to ipv6, which wasmd doesn't support

packages-ts/gauntlet-cosmos-contracts/package.json
```diff
@@ -21,7 +21,10 @@
     "preinstall": "node ../../scripts/require-yarn.js",
     "gauntlet": "ts-node ./src/index.ts",
     "lint": "tsc",
-    "test": "SKIP_PROMPTS=true jest --runInBand",
+    "dev:test:wasmd-up": "yarn ts-node ./test/scripts/startWasmdAndUpload.ts",
```
augustbleeds: my dev workflow is to run `dev:test:wasmd-up` up, then just run `dev:test` so I don't need to start wasmd

packages-ts/gauntlet-cosmos-contracts/src/commands/contracts/access_controller/deploy.ts
```diff
@@ -0,0 +1,33 @@
+import { AbstractInstruction, instructionToCommand } from '../../abstract/executionWrapper'
```
augustbleeds, line 1: had to add deploy command since it didn't exist before. Previously ops was still able to deploy thanks to this https://github.com/smartcontractkit/chainlink-cosmos/blob/augustus.BCI-1511.more-gauntlet-tests/packages-ts/gauntlet-cosmos-contracts/src/commands/abstract/index.ts#L74 find by @cfal

packages-ts/gauntlet-cosmos-contracts/src/commands/contracts/ocr2/close/acceptProposal.ts
```diff
@@ -12,9 +12,14 @@ const makeInput = async (flags): Promise<AcceptProposalInput> => {
   }
   if (flags.input) return { ...flags.input, ...defaultInput }
 
+  if (!process.env.SECRET) {
```
augustbleeds: had to add this because I changed [packages-ts/gauntlet-cosmos-contracts/src/commands/contracts/ocr2/proposal/acceptProposal.ts](https://github.com/smartcontractkit/chainlink-cosmos/pull/364/files#diff-4adab7e87a1782cdf7257c950b247df53842ede9b054aba954939b83c7aad5b2)

packages-ts/gauntlet-cosmos-contracts/src/commands/contracts/ocr2/proposal/acceptProposal.ts
```diff
@@ -44,7 +45,7 @@ const validateRandomSecret: ValidateFn<CommandInput> = async (input) => {
 const validateOffchainConfig: ValidateFn<CommandInput> = async (input, context) => {
   const { offchainConfig } = await serializeOffchainConfig(
     input.offchainConfig,
-    process.env.SECRET!,
```
augustbleeds, line 47: get from the input rather than the env so it can be passed into flags if necessary

packages-ts/gauntlet-cosmos-contracts/src/commands/contracts/ocr2/withdrawFunds.ts
```diff
@@ -30,7 +30,7 @@ const makeCommandInput = async (flags: any, args: string[]): Promise<CommandInpu
 
 const makeContractInput = async (input: CommandInput, context: ExecutionContext): Promise<ContractInput> => {
   const queryAvailableLink = async () =>
-    ((await context.provider.queryContractSmart(context.contract, 'link_available_for_payment' as any)) as any).amount
+    ((await context.provider.queryContractSmart(context.contract, {link_available_for_payment: {}} as any)) as any).amount
```
augustbleeds: a result of changing the query msg type from string to object with a string key (see first bullet point here for more context https://github.com/smartcontractkit/chainlink-cosmos/pull/359)

packages-ts/gauntlet-cosmos-contracts/src/commands/contracts/ocr2/withdrawPayment.ts
```diff
@@ -48,14 +47,13 @@ const afterExecute = () => async (response: Result<TransactionResponse>) => {
     logger.error('Could not retrieve events from tx')
     return
   }
+  const paidEvent =  events.find(e => e['type'] as any === 'wasm-set_config')
```
augustbleeds: refactor the method so it doesn't error if balance due to payee is 0

packages-ts/gauntlet-cosmos-contracts/src/commands/contracts/proxy_ocr2/confirmContract.ts
```diff
@@ -13,10 +13,8 @@ type ContractInput = {
 }
 
 const makeCommandInput = async (flags: any, args: string[]): Promise<CommandInput> => {
-  const contract = args[0]
```
augustbleeds, line 16: this was a bug because the arguments to the confirm and propose method are not the proxy contract address but the contract address

packages-ts/gauntlet-cosmos-contracts/src/lib/constants.ts
```diff
@@ -43,8 +43,8 @@ export const TOKEN_UNIT = new BN(10).pow(new BN(TOKEN_DECIMALS))
 export const UATOM_DECIMALS = 6
 
 export const EMPTY_TRANSMITTERS = [
-  'cosmos1deadg6umpcwdr5ayj737vt49agnpxc6e5ct65t',
-  'cosmos1deadmlvdvtnnclqvdt526fz465w7v3vkd5r3h7',
-  'cosmos1deadz8q0v8yf4vw6jtnwfush2yka230jggws9p',
-  'cosmos1deadxqgqq63t9hxragwcqd2h020m2hvt3l49va',
+  'wasm1hft9sxhx7d7furw9y0rjxu4hfsm76ehkman78g',
```
augustbleeds, line 46: wasmd instead of cosmos bech32 prefix

scripts/wasmd.sh
```diff
@@ -4,7 +4,7 @@ container_name="chainlink-cosmos.wasmd"
 container_version="v0.40.1"
 genesis_account="wasm1lsagfzrm4gz28he4wunt63sts5xzmczwda8vl6"
 
-set -euo pipefail
+set -euox pipefail
```
cfal, line 7: debugging change?

wasmd.nix
```diff
@@ -4,7 +4,7 @@
 
 stdenv.mkDerivation rec {
   name = "wasmd-${version}";
-  version = "0.3.0";
+  version = "0.40.1";
```
cfal, line 7: good catch!

packages-ts/gauntlet-cosmos-contracts/src/commands/contracts/access_controller/deploy.ts
cfal: if we have all specific deploy commands now, could we remove the abstract code then?

packages-ts/gauntlet-cosmos-contracts/src/commands/contracts/ocr2/proposal/acceptProposal.ts
cfal: should we still accept process.env.SECRET? or consistently use it/not use it across packages?

i'm looking at proposeOffchainConfig, which still uses it: https://github.com/smartcontractkit/chainlink-cosmos/blob/develop/packages-ts/gauntlet-cosmos-contracts/src/commands/contracts/ocr2/proposeOffchainConfig.ts#L112

in fact, it seems to require it because this check wasn't removed when adding the flag option: https://github.com/smartcontractkit/chainlink-cosmos/blob/develop/packages-ts/gauntlet-cosmos-contracts/src/commands/contracts/ocr2/proposeOffchainConfig.ts#L80 - maybe we can fix

packages-ts/gauntlet-cosmos-contracts/src/commands/contracts/ocr2/withdrawPayment.ts
```diff
@@ -48,14 +47,13 @@ const afterExecute = () => async (response: Result<TransactionResponse>) => {
     logger.error('Could not retrieve events from tx')
     return
   }
+  const paidEvent = events.find((e) => (e['type'] as any) === 'wasm-set_config')
```
cfal: why is a set_config event the paid event?

packages-ts/gauntlet-cosmos-contracts/src/lib/constants.ts
```diff
@@ -43,8 +43,8 @@ export const TOKEN_UNIT = new BN(10).pow(new BN(TOKEN_DECIMALS))
 export const UATOM_DECIMALS = 6
```
cfal: since you're in the vicinity, UATOM_DECIMALS to UCOSM_DECIMALS?

cfal (commented):
looks great, thanks for taking the time. left some comments

packages-ts/gauntlet-cosmos-contracts/src/lib/constants.ts
calvwang9: are these randomly generated wasm-prefixed addresses? I noticed the cosmos-prefixed ones start with `dead` to make it clear that they are burn addresses/empty transmitters - are we able to do something similar?

packages-ts/gauntlet-cosmos-contracts/test/utils.ts
```diff
@@ -0,0 +1,201 @@
...
+
+/**
+ * Initializes Wasmd and Contracts, unless SKIP_WASMD_SETUP=true
+ * Can save time for debugging / local testing if Wasmd and base contracts have been setup already
+ *
+ * Will read initialized account addresses from local file system if SKIP_WASMD_SETUP=true
+ *
+ * @returns {string[]} Initialized account addresses
+ */
+export const maybeInitWasmd = async () => {
```
calvwang9: copied from the other closed PR: 'maybeInitWasmd' seems kinda confusing semantically (is wasmd initialized or not?)- perhaps just 'initWasmd' would suffice?

calvwang9 (commented):
Am curious how long this the whole gauntlet test suite takes to run, as it looks like we need to tear down/spin up and upload contracts for wasmd for each test file 🤔 If it does take a while, perhaps we can explore using templates to save the 'base state' for resetting to help speed things up - https://github.com/CosmWasm/wasmd/blob/main/README.md#ci

packages-ts/gauntlet-cosmos-contracts/src/commands/contracts/access_controller/deploy.ts
augustbleeds: We can't remove the AbstractInstructions or AbstractCommand. They are inherited by every command (in this case they are called abstract in the OOP-sense not because they are meant for deploy/execution/query commands with no gauntlet command implementation but because concrete gauntlet commands use their abstract methods)

packages-ts/gauntlet-cosmos-contracts/src/commands/contracts/ocr2/proposal/acceptProposal.ts
augustbleeds: Good point! Yeah that sounds more flexible to me so that engops can either invoke it as `yarn gauntlet .... --secret $ANYSECRETNAME ... ` instead of relying on a specific environment variable to be set.

packages-ts/gauntlet-cosmos-contracts/src/lib/constants.ts
augustbleeds: oo good point. I don't know how they did that but will look into it

augustbleeds: Yeah these are just random addresses I generated with a script

scripts/wasmd.sh
augustbleeds: I like -x by default so you can follow what command triggered a failure in case one happens. But I can remove if it it's too noisy

packages-ts/gauntlet-cosmos-contracts/src/commands/contracts/ocr2/withdrawPayment.ts
augustbleeds: I think this is a typo, good eyes 👁️ 👁️

packages-ts/gauntlet-cosmos-contracts/src/commands/contracts/access_controller/deploy.ts
cfal: maybe i'm misunderstanding - but since you have specific deploy commands now, and the clause linked in the index.ts file above was previously handling `deploy` when there was no specific deploy command implemented, then can't that clause be removed?

packages-ts/gauntlet-cosmos-contracts/src/commands/contracts/ocr2/close/acceptProposal.ts
cfal: should this also accept the secret flag then?

augustbleeds:
> Am curious how long this the whole gauntlet test suite takes to run, as it looks like we need to tear down/spin up and upload contracts for wasmd for each test file 🤔 If it does take a while, perhaps we can explore using templates to save the 'base state' for resetting to help speed things up - https://github.com/CosmWasm/wasmd/blob/main/README.md#ci

Cool that's pretty neat! Currently all tests run at around 5 minutes on CI so I think it's sufficient for now. Using the lite mode we only need to setup/teardown once (which takes around 30 seconds). I think the actual uploading contracts part only takes like < 10 seconds of the 30 seconds. The majority of the time is spent actually running contract tests. 

I think that the optimization would work, but it would only save us like 15 seconds (as of today). Might be worth revisiting in the future though.

packages-ts/gauntlet-cosmos-contracts/test/utils.ts
augustbleeds: Sounds good, I can change it. I originally but "maybe" because if SKIP_WASMD=true then wasmd is not initialized. But it might not be clear from the naming

cfal (approved):

PR #365, opened 2023-08-31
integration-tests: convert cosmos pub keys to prefixed addresses

cfal:
BCI-1740

### DIFF
--- a/pkg/cosmos/params/utils.go
+++ b/pkg/cosmos/params/utils.go
@@ -1,0 +1,35 @@
+package params
+
+import (
+	"crypto/sha256"
+	"encoding/hex"
+	"errors"
+
+	"github.com/cometbft/cometbft/crypto"
+	"github.com/cosmos/cosmos-sdk/crypto/keys/secp256k1"
+	"github.com/cosmos/cosmos-sdk/types/bech32"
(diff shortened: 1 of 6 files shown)


pkg/cosmos/params/utils.go
```diff
@@ -0,0 +1,35 @@
...
+	"errors"
+
+	"github.com/cometbft/cometbft/crypto"
+	"github.com/cosmos/cosmos-sdk/crypto/keys/secp256k1"
+	"github.com/cosmos/cosmos-sdk/types/bech32"
+	"golang.org/x/crypto/ripemd160" //nolint: staticcheck
+)
+
+// Creates a bech32 address from a hex-encoded secp256k1 public key.
+func CreateBech32Address(pubKey, accountPrefix string) (string, error) {
```
calvwang9, line 15: not a big deal but iirc, this function/logic is duplicated in the keystore in core right? is it worth consolidating into one function across both repos?

calvwang9 (approved):

pkg/cosmos/params/utils.go
cfal: definitely - i'm planning on removing the keystore code and get it to use this function.

PR #367, opened 2023-08-31
BCI-1511: Fix Gauntlet Upload -- update schema dependency

augustbleeds:
fixes bug where `yarn gauntlet upload` could not find the name of the json message apis 

Was kind of a tricky bug to track down but it turns out that this branch was using an outdated version of cosmwasm-schema crate which didn't output some of the schema files. So when make contracts_compile was ran, the schema files were regenerated incorrectly. 

Some other stuff was in a broken state so I had to pick out other schema / go-related / ts changes from https://github.com/smartcontractkit/chainlink-cosmos/pull/364  that are either necessary for this PR to run successfully

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -221,9 +221,9 @@ dependencies = [
 
 [[package]]
 name = "cosmwasm-schema"
-version = "1.1.5"
+version = "1.3.3"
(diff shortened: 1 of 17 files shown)


calvwang9 (approved):
nice catch!

cfal:
thanks for the quick fix!

PR #370, opened 2023-09-07
BCI-1469: Update CosmWasm contracts

calvwang9:
https://smartcontract-it.atlassian.net/browse/BCI-1469

Update CW Plus contract versions to v1.1.0

Reviewers can ignore changes to the artifacts folders, which are downloaded from https://github.com/CosmWasm/cw-plus/releases/tag/v1.1.0


integration-tests/common/common.go
```diff
@@ -127,6 +127,8 @@ func NewCommon(t *testing.T) *Common {
 	chainlinkConfig := fmt.Sprintf(`[[Cosmos]]
 Enabled = true
 ChainID = '%s'
+Bech32Prefix = 'wasm'
```
cfal: good catch!

packages-ts/gauntlet-cosmos-contracts/src/lib/contracts.ts
```diff
@@ -95,6 +95,15 @@ export abstract class Contract {
     const abi = possibleContractPaths
       .filter((path) => existsSync(`${path}/${this.dirName}/schema`))
       .map((contractPath) => {
+        if (existsSync(path.join(contractPath, `./${this.dirName}/schema/${this.dirName.replace('_', '-')}`))) {
```
cfal: this might be outside the scope of this PR and maybe a question for @augustbleeds, but i'm curious why in the repo, we have: `contracts/ocr2/schema/main/ocr2.json` with the `main` subfolder - but then `contracts/proxy-ocr2/schema/proxy-ocr2.json` - which fits into this pattern being added.

also, with this new clause, it seems like we don't need the duplicated execute/query/instantiate files. could those be excluded from the repo or build?

eg `contracts/ocr2/schema/raw/query.json`

.. at the same time, there exists `contracts/flags/schema/raw/query.json` but no equivalent `contracts/flags/schema/flags.json` - could we be consistent here?

packages-ts/gauntlet-cosmos-contracts/src/lib/constants.ts
```diff
@@ -15,7 +15,7 @@ export const enum CATEGORIES {
 }
 
 export const DEFAULT_RELEASE_VERSION = 'local'
-export const DEFAULT_CWPLUS_VERSION = 'v0.16.0'
+export const DEFAULT_CWPLUS_VERSION = 'v1.1.0'
```
cfal, line 18: there's a single invocation of CosmWasmContract afaict: https://github.com/smartcontractkit/chainlink-cosmos/blob/fd629af1aa16469a0bda9bce8221cb6e3be1c9a4/packages-ts/gauntlet-cosmos-contracts/src/lib/contracts.ts#L156

which does not provide the `defaultVersion` (3rd) parameter at all:

https://github.com/smartcontractkit/chainlink-cosmos/blob/fd629af1aa16469a0bda9bce8221cb6e3be1c9a4/packages-ts/gauntlet-cosmos-contracts/src/lib/contracts.ts#L124

which means that version is always `v1.1.0` and never `local`, and this clause never gets invoked for CWPLUS i think? https://github.com/smartcontractkit/chainlink-cosmos/blob/fd629af1aa16469a0bda9bce8221cb6e3be1c9a4/packages-ts/gauntlet-cosmos-contracts/src/lib/contracts.ts#L58

in which case, if it always gets downloaded, i wonder why we need to download it in the Makefile changes above? could we simply remove it? or, add it to our repo locally and remove the js clause so we don't depend on a download from github during deploy? wdyt?

https://github.com/smartcontractkit/chainlink-cosmos/blob/fd629af1aa16469a0bda9bce8221cb6e3be1c9a4/Makefile#L74-L79

packages-ts/gauntlet-cosmos-cw-plus/artifacts/contracts/cw20_base/schema/cw20-base.json
```diff
@@ -1,6 +1,6 @@
 {
   "contract_name": "cw20-base",
-  "contract_version": "0.16.0",
+  "contract_version": "1.1.0",
```
cfal, line 3: is it strange that we download the contract each time, but we have the contract schema locally? 😄

packages-ts/gauntlet-cosmos-contracts/src/lib/contracts.ts
augustbleeds: >  why in the repo, we have: contracts/ocr2/schema/main/ocr2.json with the main subfolder - but then contracts/proxy-ocr2/schema/proxy-ocr2.json - which fits into this pattern being added.

For some reason, codegen did not work correctly if there was any other files besides the full abi .json and the raw/ folder OR just the full abi .json. I'm assuming this is because it expects the updated write_api! macro to be used when generating the abis. So weird behavior happens when you include additional abis on top of it. So you'll notice [here](https://github.com/smartcontractkit/chainlink-cosmos/blob/develop/packages-ts/gauntlet-cosmos-contracts/src/scripts/codegen.js) that the deviationflaggingvalidator and ocr2 contracts have this special path.

Yes, that's a good point! I kept the old abis in case we used them elsewhere and was planning to delete them after I finished the gauntlet work and found that they could be deleted. They can definitely be deleted (modifying the schema.rs to just use the write_api!) and that would remove the extra "main" folder (just remember to modify the codegen.js file too), and this new logic calvin has to pull the execute, instantiate, and query abis out of there for all contracts. That'd be awesome.

calvwang9: Thanks for the suggestions and clarifications! I've cleaned it up so that we only generate and use the full abi json 👍

packages-ts/gauntlet-cosmos-contracts/src/lib/constants.ts
calvwang9: good catch! you're right it seems to be unused - i've removed the curl in the makefile (tests still run fine)

packages-ts/gauntlet-cosmos-cw-plus/artifacts/contracts/cw20_base/schema/cw20-base.json
calvwang9: thats a good point actually - changed loadContractABI to fetch the abi from github as well. We can now remove all the local artifacts related to cw+ contracts

calvwang9: actually - had to restore the locally stored artifacts as it looks like codegen.js in the gauntlet tests rely on these

contracts/deviation-flagging-validator/examples/schema.rs
```diff
@@ -18,22 +18,4 @@ fn main() {
         execute: ExecuteMsg,
         query: QueryMsg,
     }
-
```
augustbleeds, line 21: 🥳

augustbleeds (approved):

cfal (approved):

PR #380, opened 2023-10-27
pkg/cosmos/config: mv config from core

jmank88:
Move config from core, which requires bumping chainlink-relay, which comes with Go 1.21.

Supports:
- https://github.com/smartcontractkit/chainlink/pull/11101


krehermann (approved):

aalu1418 (approved):
approving + merging to unblock core extraction - broken gauntlet + contract tests will be revisited in the future

PR #381, opened 2023-10-31
pkg/cosmos: move tx and chain from core

jmank88:
https://smartcontract-it.atlassian.net/browse/BCI-900

Supports:
- https://github.com/smartcontractkit/chainlink/pull/11136


pkg/cosmos/txm/orm.go
```diff
@@ -0,0 +1,133 @@
...
+	"fmt"
+
+	"github.com/smartcontractkit/sqlx"
+
+	"github.com/smartcontractkit/chainlink-cosmos/pkg/cosmos/adapters"
+	"github.com/smartcontractkit/chainlink-cosmos/pkg/cosmos/db"
+)
+
+type DB interface {
+	BeginTxx(context.Context, *sql.TxOptions) (*sqlx.Tx, error)
```
cfal, line 16: BeginTx?

jmank88: This is the sqlx variant that we use in core. IIUC they couldn't override `BeginTx` while returning a different `Tx` type, so they suffixed with `x`. Bad luck that `Tx` also ends in `x` in this case :shrug:

cfal (approved):

krehermann (approved):

aalu1418 (approved):

aalu1418:
merging to unblock cosmos extraction work in core - CI will continue to fail

PR #413, opened 2024-08-06
use gomods; add make generate; add CI check for tidiness+generation

jmank88:
This PR swaps in [gomods](https://github.com/jmank88/gomods) for tidy and generate commands, updates stale generated files, and adds a CI check for tidiness+generation.


.github/workflows/relay.yml
jmank88, line 1: This should be required or folded in to a required job.

aalu1418: made it required!

aalu1418 (approved):

PR #438, opened 2025-04-10
bumping common

patrickhuie19:
core ref: 37b745429f597a748ad4299c47530dac4fbf92d9


integration-tests/ocr2_test.go
```diff
@@ -62,8 +63,8 @@ func TestOCRBasic(t *testing.T) {
 	require.NoError(t, err, "Could not create private key from mnemonic")
 	logger.Info().Str("from", testAccount.String()).Msg("Funding nodes")
 
-	gasPrice := types.NewDecCoinFromDec("ucosm", types.MustNewDecFromStr("1"))
-	amount := []types.Coin{types.NewCoin("ucosm", types.NewInt(int64(10000000)))}
+	gasPrice := types.NewDecCoin("ucosm", math.NewInt(int64(10000000)))
```
jmank88: @EasterTheBunny the old apis are available with `Legacy` prefixes if that keeps this more readable.

EasterTheBunny: Ah ok. I just did a quick fix to get it working, but I can use the legacy stuff

EasterTheBunny: It looks like my change fixed the interface issues, but made the test not pass. I'll update it and push a new commit.

